PR #1, opened 2021-01-21
Initial file structure commit

jaretanderson:
If you like this file structure then we can go ahead and merge it over to main. This has a *main.cpp* file which must be compiled with all of the utility scripts in the *util* directory inside FSW. Function declarations should go in the corresponding header file with the same name as the util script, and that header should be included in the main script as well as the util script it corresponds to according to standard C++ development practices, but I can explain this more if you aren't familiar with it. If you don't have any feedback, you can go ahead and keep using the comment block format that I have in my files in any new helper scripts you create. 

I am hoping that we can do most of our development in our own util scripts so that we don't have to worry about merge conflicts too often, but from an organizational standpoint it makes sense to develop similar modules within the same util script so that we can reuse functions within that util script across multiple modules.  

Eventually, we can do our development for GSW in the *GSW* dir. I have also created an *Arduino* dir for when we get the FSW to a state that we want to convert it to Arduino sketches, although I am not super sure about the specifics of it yet. For now I am compiling using g++, but I am worried that we might have to change this because I read that the Arduino IDE uses avr-gcc. We will have to do some hardware tests before I am super confident about this. 

Jaret


FSW/headers/memUtil.hpp
```diff
@@ -0,0 +1,17 @@
+#ifndef MEMUTIL_H
```
JashanChopra, line 1: Are these header files just for convivence of not included all the declarations/definitions/libraries in the actual .cpp files?

FSW/util/memUtil.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * Outputs:
+ *  None
+ */
+void dataProcessing()
+{
+    std::cout << "Data Processing Module: Test Print" << std::endl;
+}
+
+
+/* - - - - - - scienceMemoryHandling - - - - - - *
```
JashanChopra, line 42: Big fan of these function headers, good for code review and important to be able to quickly glance and what functions in main do

JashanChopra: I tend to put them inside the function definition but it doesn't really matter as long as it is consistent

FSW/main.cpp
```diff
@@ -0,0 +1,44 @@
...
+ *
+ */
+
+/* - - - - - - Includes - - - - - - */
+// C++ libraries
+
+// Other libraries
+
+// NS2 headers
+#include "headers/memUtil.hpp"
```
JashanChopra, line 17: I'm unfamiliar with how this inclusions work in C++. What would make sense to me is that you require the main file (memUtil.cpp) which in turn requires the header file. Can you explain this to me?

FSW/headers/memUtil.hpp
jaretanderson: In C/C++, there are two things required to successfully compile a function: the function declaration and the function definition. When using multi-file compilation (as we are), headers are a way to ensure that all of the function declarations are preprocessed before the compiler gets to any function calls inside function definitions. If the compiler arrives at a function call before it has processed the definition, then it will throw an error and fail to compile. Using `#include <header.hpp>` ensures that all function definitions are seen by the compiler before the function calls.

FSW/main.cpp
jaretanderson: the key here is that when compiling the program, we run a line like this:

`g++ script1.cpp script2.cpp ... scriptN.cpp`

This compiles all of the files (so `main.cpp` and each util script in our case) into a single executable. The header files being included ensures that all functions being called are defined properly and do not throw a compilation error. Putting each of these declarations above `main()` in `main.cpp` is tedious for large projects, which is why we use header files.

FSW/main.cpp
```diff
@@ -0,0 +1,44 @@
...
+ */
+int main()
+{
+    // TODO: Change main loop to account for critical exit conditions
+    while(true) // run main loop forever
+    {
+        // test multiple file structure by calling functions from memUtil.cpp
+        dataProcessing();
+        scienceMemoryHandling();
+        break;
```
JashanChopra, line 40: Right now it breaks upon completion for testing, but how does this loop run in reality? After a critical exit condition occurs, is there methodology to get back into the loop? Something like

```
begin
    function()
rescue
    handleException()
    continue
end
```

Or is there just a clean break and exception handling is manual, with a CMD sent to the spacecraft to then restart main.cpp

FSW/main.cpp
jaretanderson: Also the Arduino IDE does this compilation for us automatically if we provide it with a file structure it understands as far as I can tell. I will want to test this though ASAP

FSW/main.cpp
jaretanderson: @Vysares I will defer to you on this one, I just put this main loop in here for super basic testing but I agree with Jashan that this will need to be fleshed out eventually

FSW/main.cpp
JashanChopra: Interesting good to know. I imagine that command will get long then as we start to have a ton of different functions. Is there anyway that is avoided, I.E with a Makefile or something?

JashanChopra (approved):
Good principles in place, as far as I can tell ;)

FSW/main.cpp
jaretanderson: yeah we could avoid it with a makefile if we wanted to. I don't anticipate having more than 4 or 5 util scripts in FSW, but I will add this as a topic for tuesday's meeting.

Vysares (approved):
Very nice work. I think this file structure is simple, elegant, and does exactly what we need. I love the headers as well.

PR #2, opened 2021-01-28
Psuedocode for basic electronics system tests

JashanChopra:
default.cpp: test the teensy is alive
lightCollect.cpp: test photodiode data collection
resistorHeater.cpp: trigger the resistor heater
watchdog.cpp: unleash the beast

Here's the psuedocode. Needs a lot of help. I think I wrote it in Arduino form? Not sure if you want to keep it as Arduino sketches or convert it to c++. I basically just want to be able to plug in our work computer in 150 to the Teensy, launch some software, and be able to click run on these scripts. Let me know if the functionality isn't making sense. 

We can review this on Friday, 1/29 at our 10:30 AM meeting. Maybe Matt can help write some of this code. SPI connection can come later if it isn't ready by the time I want to test. 

Hoping for a ~2 week turnaround on this, if possible.

### DIFF
--- a/Test/default.cpp
+++ b/Test/default.cpp
@@ -10,19 +10,6 @@
 // Other libraries
 
 // NS2 headers
-
-/* - - - - - - Functions - - - - - - */
-
-/* - - - - - - main - - - - - - *
- * Usage:
- *  Main function entered on runtime
- * 
- * Inputs:
- *  none
- *  
- * Outputs:
- *  None
- */
  
  // I think I'm writing this as Arduino code not c++? 
 int aRegCurr = 17;
--- a/Test/default.cpp
+++ b/Test/default.cpp
@@ -1,0 +1,67 @@
+/* default.cpp verifies the Teensy is connected for NanoSAM II tests. 
+ * Usage:
+ *  run this script through Teensyduino to power the NS2 payload
+ *
+ */
+
+/* - - - - - - Includes - - - - - - */
+// C++ libraries
+
+// Other libraries
+
+// NS2 headers
+
+/* - - - - - - Functions - - - - - - */
+
+/* - - - - - - main - - - - - - *
+ * Usage:
+ *  Main function entered on runtime
+ * 
+ * Inputs:
+ *  none
+ *  
+ * Outputs:
+ *  None
+ */
+ 
+ // I think I'm writing this as Arduino code not c++? 
+int aRegCurr = 17;
+int dRegCurr = 18;
+int dRegPG = 20; 
+
+void loop()
+{
+    while(true) // run main loop forever
+    {
+		// Check voltage regulator currents
+        aCurr = analogRead(aRegCurr);
+		dCurr = analogRead(dRegCurr);
+		dPG = analogRead(dRegPG);
+		
+		
+		// Print statements for confirmation
+		// TODO: Gotta be a better way to format these statements?
+		Serial.print('Analog Current Value [A]: ');
+		Serial.print(aCurr);
+		Serial.print('\n');
+		
+		Serial.print('Digital Current Value [A]: ');
+		Serial.print(dCurr);
+		Serial.print('\n');
+		
+		Serial.print('Digital Power Good [V]: ');
+		Serial.print(dPG);
+		Serial.print('\n');
+		
+		// if the power good signal returns around 3.3V, power is good
+		// TODO: I have no clue how to write an if statement in arduino
+		if (dPG > 3.25) && (dPG < 3.35)
+			Serial.print('Good to go')
+		end
+		
+        break;
+    }
+
+    return 0;
+}
+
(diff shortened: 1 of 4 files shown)


Test/default.cpp
```diff
@@ -0,0 +1,67 @@
...
+        aCurr = analogRead(aRegCurr);
+		dCurr = analogRead(dRegCurr);
+		dPG = analogRead(dRegPG);
+		
+		
+		// Print statements for confirmation
+		// TODO: Gotta be a better way to format these statements?
+		Serial.print('Analog Current Value [A]: ');
+		Serial.print(aCurr);
+		Serial.print('\n');
```
jaretanderson, line 33: Looks like there is string concatenation using the ' + ' operator (https://www.arduino.cc/en/Tutorial/BuiltInExamples/StringAdditionOperator), however it requires relatively specific syntax to avoid operator precedence errors. Declaring these on multiple lines appears to be the safest, and it is how the example files in the SerialFlash library does it.

Test/watchdog.cpp
```diff
@@ -0,0 +1,67 @@
...
+		Serial.print('\n');
+		
+		Serial.print('Digital Power Good [V]: ');
+		Serial.print(dPG);
+		Serial.print('\n');
+		
+		// if the power good signal returns around 3.3V, power is good
+		// TODO: I have no clue how to write an if statement in arduino
+		if (dPG > 3.25) && (dPG < 3.35)
+			Serial.print('Good to go')
```
jaretanderson: I'll probably add in a line printing the expected range to avoid confusion for the user, so this means that these voltage range definitions should be done up top instead of hardcoded

Test/watchdog.cpp
```diff
@@ -11,55 +12,19 @@
...
-		if (dPG > 3.25) && (dPG < 3.35)
-			Serial.print('Good to go')
-		end
+		// Send a 3.3V, 50ns width square wave to trigger dog
+		squareWave = 0; // TOOD: Construct square wave
+		digitalWrite(dogPin, squareWave);
 		
-        break;
+		// After signal, the Teensy should have shut off
+		// TODO: Some sort of reset function? Not sure how this works...
```
jaretanderson, line 26: Do you want this reset function to power the teensy back up?

Test/watchdog.cpp
```diff
@@ -11,55 +12,19 @@
...
-		Serial.print(dPG);
-		Serial.print('\n');
-		
-		// if the power good signal returns around 3.3V, power is good
-		// TODO: I have no clue how to write an if statement in arduino
-		if (dPG > 3.25) && (dPG < 3.35)
-			Serial.print('Good to go')
-		end
+		// Send a 3.3V, 50ns width square wave to trigger dog
+		squareWave = 0; // TOOD: Construct square wave
```
jaretanderson, line 22: Doesn't the dog need to be fed to keep the system on? Wouldn't the purpose of this pulse be to prevent the watchdog from triggering, not the opposite?

jaretanderson (approved):
Overall looks very doable, I will have to learn SPI for the Science Memory handling anyway so that shouldn't be a problem to implement in the next two weeks. Will be good to get a real test with SPI in before too long. 

It is totally fine to have a bunch of different .cpp files for this, we will just have to stay organized as we create more tests for different subteams/system level tests.

Test/watchdog.cpp
JashanChopra: That got totally flipped up in my mind, you are correct. Sorry. A 50ns width square wave must be sent to pin 2 every 1.2 seconds or the voltage input to the Teensy will be shorted. 

So we actually need to develop a watchdog module and incorporate it into all the other tests.

This watchdog test can become a "reset" test. This should be able to be accomplished by a long enough pause. I'd like to see the software reset and enter a loop again, unless our plan is to manually trigger a "turn on" function after reset.

Test/watchdog.cpp
JashanChopra: Re: other message. Yes, unless GSW design has different plans.

Test/default.cpp
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+/* - - - - - - Includes - - - - - - */
+// C++ libraries
+
+// Other libraries
+
+// NS2 headers
+ 
+ // I think I'm writing this as Arduino code not c++?
```
Vysares, line 14: Arduino code is C++, just with the inclusion of the Arduino library and the caveat that not all C++ libraries work on our hardware.

Test/watchdog.cpp
Vysares: Yes that's how a watchdog works. If it doesn't get fed it triggers a hard reset without consulting the software. I believe we have it configured on the board to just cycle the power but I'm not sure about that. @JashanChopra ?

Vysares (approved):

PR #4, opened 2021-01-29
Simplified NanoSAM 1 testing software

Vysares:
Gutted the old NS1 code to continuously read the ADC and transmit each value via serial. Its functionality is dubious at best, but it compiles and runs fine on my teensy, returning lots of zeros as expected.

### DIFF
--- a/Test/testNS1.cpp
+++ b/Test/testNS1.cpp
@@ -1,0 +1,45 @@
+/* testNS1.cpp continuously reads photodiode data and transmits each data point via serial
+ * Usage:
+ *  This is a quick and dirty script based on the original NS1 software to test the basic functionality of NanoSam I
+ *  Run this script through Teensyduino to test the NS1 payload, monitor via the serial monitor/plot
+ *  
+ */
+
+/* - - - - - - Includes - - - - - - */
+// C++ libraries
+
+// Other libraries
+#include <SPI.h>
+
+/* - - - - - - Initialization - - - - - - */
+const int ADC_CHIP_SELECT = 10; //chip select pin number for ADC
+
+/* - - - - - - Functions - - - - - - */
+
+// getADC reads and returns one 16 bit data point from the ADC via SPI
+uint16_t getADC()
+{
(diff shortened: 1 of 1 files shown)


Test/testNS1.cpp
```diff
@@ -0,0 +1,45 @@
...
+// Other libraries
+#include <SPI.h>
+
+/* - - - - - - Initialization - - - - - - */
+const int ADC_CHIP_SELECT = 10; //chip select pin number for ADC
+
+/* - - - - - - Functions - - - - - - */
+
+// getADC reads and returns one 16 bit data point from the ADC via SPI
+uint16_t getADC()
```
jaretanderson, line 21: do we always have to use this custom defined uint16_t instead of float when dealing with the ADC? more to know for myself in the future

jaretanderson (approved):

Test/testNS1.cpp
Vysares: I'm not sure, both are 16 bits so I would assume both are acceptable, if for some reason we have to use uint16 to read the SPI we can convert it to a float to work with the code you already wrote.

Test/testNS1.cpp
```diff
@@ -0,0 +1,45 @@
...
+{
+  uint16_t photodiode16;
+  SPI.beginTransaction(SPISettings(50, MSBFIRST, SPI_MODE3)); //SPISettings(maxSpeed,dataOrder,dataMode)
```
JashanChopra, line 24: Why this dataOrder and dataMode? I have no qualms - just curious what the options are.

JashanChopra (approved):

Test/testNS1.cpp
Vysares: All three of those arguments are dictated by the slave — the ADC in this case, it may be different for the flash modules. I just copied it over from the old code so I don't know if they got it right. I am particularly doubtful of that maxSpeed since it is usually in the MHz range.

PR #5, opened 2021-02-02
Converted EPS testing pseudocode

Vysares:
Turned Jashan's pseudocode into real code. Watchdog functionality included.  I kept these testing scripts in one file each so we can just throw it into an arduino sketch and upload it, and so it is easy to make edits during testing if we need to. For the actual FSW the watchdog feeding will be handled in essentially the same way, but in one of the util files.

@JashanChopra Please let me know if I messed up any pin numbers.

### DIFF
--- a/FSW/headers/eventUtil.hpp
+++ b/FSW/headers/eventUtil.hpp
@@ -1,0 +1,52 @@
+#ifndef EVENTUTIL_H
+#define EVENTUTIL_H 
+
+/* - - - - - - Includes - - - - - - */
+// C++ libraries
+#include <vector>
+
+// Other libraries
+#include <Arduino.h>
+
+// NS2 headers
+
+
+/* - - - - - - Declarations - - - - - - */
+
+/* - - - - - - Class Definitions - - - - - - */
+
+class Event
+{
+    protected:
+        bool isInvoked;
+        
+    public:
+        Event();
+        bool checkInvoked();
+        void invoke();
+};
+
+
+class TimedEvent : public Event
+{
+    protected:
+        unsigned long duration;
+        unsigned long currentMillis;
+        unsigned long nextInvokeMillis;
+
+    public:
+        TimedEvent(unsigned long newDuration);
+        void setDuration(unsigned long newDuration);
+        bool checkInvoked();
+        void start();
+};
+
+class RecurringEvent : public TimedEvent
+{
+    public:
+        RecurringEvent(unsigned long newDuration);
+        bool checkInvoked();
+};
+
+
+#endif
--- a/Test/default.cpp
+++ b/Test/default.cpp
@@ -26,16 +26,16 @@ const float VMAX = 3.35;        // maximum acceptable voltage
 
 // timing
 const int SAMPLE_INTERVAL = 3000;       // time between samples in ms
-const int WD_RESET_INTERVAL = 1000;     // watchdog feeding interval, ms
+const int WD_RESET_INTERVAL = 100;     // watchdog feeding interval, ms
 const int WD_PULSE_DUR = 10;            // watchdog reset signal duration, MICROSECONDS!!!
 
 int wdLastFeedMillis;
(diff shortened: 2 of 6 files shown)


FSW/headers/eventUtil.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <Arduino.h>
+
+// NS2 headers
+
+
+/* - - - - - - Declarations - - - - - - */
+
+/* - - - - - - Class Definitions - - - - - - */
+
+class Event
```
JashanChopra, line 18: It would help if we write up either some external documentation or write some general code comments in here about what the classes do. I think classes can get confusing when there ends up being a lot to go track.

Test/default.cpp
```diff
@@ -8,47 +8,108 @@
...
- // I think I'm writing this as Arduino code not c++? 
-int aRegCurr = 17;
-int dRegCurr = 18;
-int dRegPG = 20; 
 
-void loop()
+
+/* - - - - - - Initialization - - - - - - */
+// pin numbers
+const int PIN_AREG_CURR = 17;   // analog regulator current pin
```
JashanChopra, line 18: Pins good

Test/default.cpp
```diff
@@ -8,47 +8,108 @@
...
+const float HI_VOLTAGE = 3.3; // high board voltage, 3.3V for teensy
+const float R_SENSE = 0.02;    // current sensor resistance, ohms
+
+
+/* - - - - - - Functions - - - - - - */
+
+/* - - - - - printStatus - - - - - */
+void printStatus()
+{
+  float aCurr = HI_VOLTAGE*(float)analogRead(PIN_AREG_CURR)/(ADC_BINS*R_SENSE);   // analog regulator current, Amps
```
JashanChopra: I think the bin conversion (3.3 / 1024) should be a constant for clarity.

JashanChopra: Also, the current (both analog and digital) is actually the same as the voltage reading of the pin, because the voltage is in units of 1 V/A. I did the math to show Jackson, but I can also show you later Jaret. I'll manually determine any error in the lab.

Test/default.cpp
```diff
@@ -8,47 +8,108 @@
...
 }
 
+/* - - - - - Setup - - - - - */
+void setup()
+{
+  // set pin modes
+  pinMode(PIN_AREG_CURR, INPUT);
+  pinMode(PIN_DREG_CURR, INPUT);
+  pinMode(PIN_DREG_PG, INPUT);
+  pinMode(PIN_WD_RESET,OUTPUT);
```
JashanChopra: Space after comma fit formatting above.

Test/lightCollect.cpp
```diff
@@ -1,49 +1,105 @@
...
 // Other libraries
+#include <Arduino.h>
+#include <SPI.h>
 
 // NS2 headers
 
+/* - - - - - - Initialization - - - - - - */
+// pin numbers
+const int PIN_ADC_CS = 10;      // ADC chip select pin
+const int PIN_PHOTO = 14;       // direct photodiode data pin for use with Teensy ADC
```
JashanChopra: PIN_PHOTO should be 21.

Test/lightCollect.cpp
```diff
@@ -1,49 +1,105 @@
...
- *  
- * Outputs:
- *  None
- */
- 
-int sciData = 14;
+/* - - - - - readDataBackup - - - - - */
+void readDataBackup() // read data direct from pin using Teensy ADC
+{
+    double sciData = HI_VOLTAGE*(float)analogRead(PIN_PHOTO)/ADC_BINS;
```
JashanChopra: Convert HI_VOLTAGE / ADC_BINS to a constant here as well.

jaretanderson (approved):

JashanChopra (approved):
few small changes necessary, looks excellent

PR #6, opened 2021-02-02
Science memory handling and data processing modules

jaretanderson:
Science Memory Handling and Data Processing modules initial commit. These compile on the Arduino IDE except for two lines dealing with the timing class. We will have to instantiate timing class objects with a scope that they can be used within `memUtil.cpp`. 

Data Processing module is actually called within Science Memory Handling, I do not think we will want to call `dataProcessing()` from main to keep the photodiode pipeline simple and memory usage (for science buffer) efficient. 

Also updated file structure to match Arduino IDE requirements for compilation (nested all additional files in `src` directory).
`SerialFlash` capabilities as well as `SPI` are untested and almost certainly will require debugging.


FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+ * Additional files needed for compilation:
+ *  none
+ */
+
+/* - - - - - - Includes - - - - - - */
+// All libraries are put in memUtil.hpp
+// NS2 headers
+#include "../headers/memUtil.hpp"
+
+/* Module Variable Definitions */
```
Vysares, line 22: We will eventually want to define these constants in main.cpp like we discussed

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+ *  data from the ADC (voltage measurement)
+ */
+float dataProcessing()
+{
+    float voltage;
+
+    // begin an SPI connection
+    // TODO: we should probably begin the SPI connection in main to avoid duplicates
+    SPI.begin();  
+    pinMode(ADC_CHIP_SELECT, OUTPUT); // set ADC chip select pin to output
```
Vysares: All the pin modes should be set in main.cpp as well, since I don't think there is ever a reason to change them.

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+ * Inputs:
+ *  none
+ *  
+ * Outputs:
+ *  None
+ */
+void scienceMemoryHandling()
+{    
+    if (dataProcessEvent.checkInvoked()){ // checks event status from timing module
+        float currentPhotodiodeVoltage = dataProcessing();
```
Vysares: Maybe rename this to just 'photodiodeVoltage'?
It's obvious what this variable is, but I think having 'current' and 'voltage' in the same variable name is a bit confusing

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+void updateBuffer(float sample, int &index){
+    
+    // check that index is valid
+    if (0 <= index && index < BUFFERSIZE){
+        dataBuffer[index] = sample;
+        index++;
+    
+    } else {
+        Serial.print("WARNING: invalid dataBuffer index ");
+        Serial.print("(Science Memory Handling Module)\n");
```
Vysares: This is great! Indicating the module that prints a warning makes debugging and testing much easier.

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    return status; // return whether or not file was successfully created
+}
+
+/* - - - - - - fullErase - - - - - - *
+ * Belongs to Science Memory Handling Module
+ *  NOT CURRENTLY SUPPORTED, chip-wide .erase() function not compiling
+ * Usage:
+ *  erases all contents of a flash module chip
+ *  this function will interrupt processing for a few minutes, only use
```
Vysares: We can't interrupt processing for more than 1.2 seconds without triggering the watchdog, so we would need to somehow cut power to the watchdog but not the flash memory to do this

Vysares (approved):
Looks good.  I am also doubtful that the SPI transactions will work on the first try — just have to test it out and work from there.

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+ *  none
+ */
+
+/* - - - - - - Includes - - - - - - */
+// All libraries are put in memUtil.hpp
+// NS2 headers
+#include "../headers/memUtil.hpp"
+
+/* Module Variable Definitions */
+//TODO: should sampling rate and window length be configurable on flight?
```
JashanChopra: I think both of these should be configurable on flight. We may want to increase sampling rate if we determine that we have the ability to average more data. We may want to extend the buffer length to capture additional data to adjust calibration times.

FSW/src/util/memUtil.cpp
JashanChopra: You would be correct. The CS pins are only ever pulled low (output). The MOSI lines are only ever master out (output) and the MISO lines are only ever master in (input)

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+    // begin an SPI connection
+    // TODO: we should probably begin the SPI connection in main to avoid duplicates
+    SPI.begin();  
+    pinMode(ADC_CHIP_SELECT, OUTPUT); // set ADC chip select pin to output
+
+    // access ADC Pin (SPI)
+    uint16_t photodiode16; 
+    SPI.beginTransaction(SPISettings(SPI_MAX_SPEED, MSBFIRST, SPI_MODE3)); //SPISettings(maxSpeed,dataOrder,dataMode)
+    digitalWrite(ADC_CHIP_SELECT, LOW);   // set Slave Select pin to low to select chip
+    photodiode16 = SPI.transfer16(0x0000);// transfer data, send 0 to slave, recieve data from ADC
```
JashanChopra, line 62: Is the "16" in reference to the number of bits?

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+    digitalWrite(ADC_CHIP_SELECT, HIGH);  // set Slave Select pin to high to de-select chip
+    SPI.endTransaction();
+
+    // Convert ADC output to voltage by scaling over voltage range
+
+    //TODO: can we scale this by the actual digital voltage on the board?
```
JashanChopra: What do you mean by "actual digital voltage on the board?"

FSW/src/util/memUtil.cpp
JashanChopra: Agreed

FSW/src/util/memUtil.cpp
JashanChopra: Yea I agree, I think you two should use this pattern elsewhere. Good example to follow with temperature checks and stuff. I think FSW should have automated triggers to enter safe modes, extending the functionality of these check blocks.

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+ *  file creation status
+ */
+bool saveBuffer(int &index){
+    
+    // create array to hold data in time ascending order
+    float timeSortArray[BUFFERSIZE];
+    int j = 0; // iterator for sorted array index
+
+    // reorder array so that it is ascending in time
+    for (int i = index; i < BUFFERSIZE; i++){
```
JashanChopra, line 157: I'm a little confused on the utility here. Why would the array of incoming data not already be ascending in time?

JashanChopra: Secondary: Even if the array was not sorted, would having the timestamp associated with the data point not make this trivial on the ground? Not sure if there's a point to expend resources on sorting here, unless needed for other applications that I am missing.

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+     * https://github.com/PaulStoffregen/SerialFlash/blob/master/README.md 
+     */
+
+    // check if file exists
+    int fileIdx; // iterator for loop checking file existence
+    bool fileFlag = true;
+    char filename[] = "scienceFile0.csv";   // null-terminated char array
+    int fileIdxOffset = 11;                 // index of file number
+
+    while (fileFlag){
```
JashanChopra, line 179: Should there be an else statement here, if the file doesn't exist, warn? Do we just lose data in this case, if the file doesn't exist and the buffer is overwritten?

FSW/src/util/memUtil.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    // write buffer to this new file
+    SerialFlashFile file;
+    file = SerialFlash.open(filename);
+    status = file.write(dataBuffer, BUFFERSIZE); //write dataBuffer
+    
+    // TODO: may need to seek position to end of file before writing again
+    //file.seek(BUFFERSIZE);  
+
+    //TODO: decide on timestamp format and append it to file
```
JashanChopra, line 208: I think Julian time would be good here, otherwise I want to say ISO 8601 is the standard.

FSW/src/util/memUtil.cpp
JashanChopra: Could you write something that manually deletes in blocks? Let me know if there's more cases of this. It might be worth a redesign effort, at least to support next year's team. 

Tangential: what scenario would require us to erase the flash chips?

JashanChopra (changes requested):
It might be a nice stopgap to always know that the code in `main` is at the least able to compile. At what point will those timing class objects be working? I think its fine to open new PRs and loop back to this and merge once it successfully compiles.

FSW/src/util/memUtil.cpp
jaretanderson: Yeah the default is to just delete single files which will work totally fine. This function is not really needed, I just saw it in the docs and figured it might be nice after testing if we generated a bunch of files and didn't want to delete them all manually. It doesn't need to go in though with that watchdog issue, I also found that the `.erase()` call was leading to compile errors suggesting the newest version of `SerialFlash.h` may not actually support it.

FSW/src/util/memUtil.cpp
jaretanderson: yeah it is to be used with the `transfer16` function in `SPI.h` but if that isn't clear I can rename this

FSW/src/util/memUtil.cpp
jaretanderson: If the voltage fluctuates at all or has an offset due to the regulator not being perfect it could affect the math a bit

e.g. bin 1024 (10 bit calculation) represents max voltage, assumed to be 3.3, but if the regulator is outputting 3.2 then our max bin should scale down to be 3.2, leading to error in the photodiode reading.

Is that true?

FSW/src/util/memUtil.cpp
jaretanderson: the buffer is continuously overwritten index-by-index and then the entire buffer is saved when a data window ends. We cannot tell when the data window starts, only when it ends on sunset. So at each timestep we overwrite the oldest value in the buffer array which is the length of one data window. When the data window ends, we know that the buffer contains the correct number of data points, but it is not in order. The most recent index overwritten is the location of the most recent value, so we need to reorder the array so that index becomes the end of the array (preserving the relative locations of the other indices). If this is still confusing I can give a better explanation with some diagrams.

To optimize memory usage we are only storing the timestamp of this most recent value and then backsolving for the remaining timestamps on the ground since we know the sampling rate and the data window is quite short.

FSW/src/util/memUtil.cpp
jaretanderson: I will add that in, good call

FSW/src/util/memUtil.cpp
JashanChopra: So the ADC is powered by a 3.3V voltage reference, so it should be quite stable over time and temperature. Your example is correct, but I think events like this are invalidated by the calibration measurement. There actually isn't a Teensy pin that reads the output of the 3.3V voltage reference, so you couldn't check the max bin size dynamically on flight.

FSW/src/util/memUtil.cpp
JashanChopra: I'm still slightly confused, but I believe you. The backsolving of the timestamps makes sense.

jaretanderson:
I agree with Jashan's comment in general that we should keep the `main` branch in a state where it compiles, but we are in a strange scenario where I need to pull main to get the event module to get my code to compile. That would be a lot easier to do if we complete the merge on my outstanding pull request. I think in general we should follow the practice of keeping main in a state where it will compile, but for this time I think it will be much easier if we don't. I can get it to a state where it will compile by the end of the day. I might also have to do some minor cleanup on the file structure, I was playing around with merging in the changes locally and running into some slight issues with the `src` reorganization.

jaretanderson:
Ok I merged in the changes from `main` and got this to a spot where it compiles. I am hoping this solved any issues with the `src` file structure change as well. We still need to clean up the actual functionality of using events to create the proper timing, and the events should probably be declared in a different location eventually.

PR #10, opened 2021-02-08
Command module dev

Vysares:
Command handling base functionality. Everything is tested on the teensy.
Added command queue size and serial timeout variables in config.hpp. 

New commands just need an entry in the Command enum (in comUtil.hpp) and an implementation in the huge switch statement (in comUtil.cpp). Commands can be easily added while keeping related commands numerically adjacent. The enum handles all the indexing for us.

Commands that affect the command handling module itself are dubbed meta-commands, they are not added to the queue, but are instead executed immediately, this way meta-commands can manipulate the behavior of the queue without affecting themselves.

### DIFF
--- a/FSW/src/util/comUtil.cpp
+++ b/FSW/src/util/comUtil.cpp
@@ -1,17 +1,16 @@
-/* hkUtil.cpp handles NS2 payload housekeeping
+/* comUtil.cpp handles NS2 payload commands
  * Usage:
- *  module functionality
- *      system state data collection
- *      temperature control
- * 
- *  function definitions
- *  put function declarations in memUtil.hpp
+ *  Reads incoming commands from serial, adds them to the command queue
+ *  and executes each command. 
+ *      
  * 
  * Modules encompassed:
- *  
+ *  Command Handling
  *
  * Additional files needed for compilation:
- *  none
+ *  config.hpp
+ *  comUtil.hpp
+ *  
  */
 
 /* - - - - - - Includes - - - - - - */
--- a/FSW/src/util/comUtil.cpp
+++ b/FSW/src/util/comUtil.cpp
@@ -22,7 +22,7 @@
 // NS2 headers
 #include "../headers/comUtil.hpp"
 
-int command[commandQueueSize];
+static CommandQueue commandQueue();
 
 
 /* - - - - - - Module Driver Functions - - - - - - */
@@ -40,28 +40,49 @@ int command[commandQueueSize];
  */
 void commandHandling()
 {
-    int newCommand = getCommand();
-    newCommand
+    int newCommandCode = getCommand();
+    
+    if (newCommandCode != -1) // if the new command is valid, add it to the queue
+    {
+        queueCommand(newCommandCode);
+    }
+
 }
 
 
-
-byte getCommand();
+int getCommand();
 {
     // read Serial
     if (Serial.available() > 0) // check if there is serial input
     {
-        int serialInput = Serial.parseInt();
-        return serialInput;
+        int serialInput = Serial.parseInt(); // read the command
+        if ((serialInput <= static_cast<int>(Command::DoNothing)) && (serialInput >= 0)) // check if command is valid
+        {
+            Serial.println("Command Recieved.")
+            return serialInput;
+        }
+        else
+        {
+            Serial.println("Invalid command recieved.");
+            Serial.println("(Command Handling)");
+        }
     }
-    return 0; // no command
+    return -1; // no command
 
 }
 
-void parseCommand(byte command)
+void queueCommand(int commandCode)
 {
-    switch newCommand
-    {
 
+}
+
+void executeCommand(int commandCode)
+{
+    switch (commandCode)
+    {
+        case EnterSafeMode:
+        {
+            
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


FSW/src/util/comUtil.cpp
```diff
@@ -40,28 +40,49 @@ int command[commandQueueSize];
...
-byte getCommand();
+int getCommand();
 {
     // read Serial
     if (Serial.available() > 0) // check if there is serial input
     {
-        int serialInput = Serial.parseInt();
-        return serialInput;
+        int serialInput = Serial.parseInt(); // read the command
+        if ((serialInput <= static_cast<int>(Command::DoNothing)) && (serialInput >= 0)) // check if command is valid
```
jaretanderson: I like this logic, however would it potentially be more useful to split up these prompts? Right now the same error is printed regardless of if the check fails due to the command code being larger than `DO_NOTHING` or if it fails due to a negative `serialInput`

Something like "Command failed due to invalid serial input" separate from "Command code outside enum bounds"

jaretanderson (approved):
Great modularized design, and I like the syntax changes of capitalizing the enum as constants.

FSW/src/util/comUtil.cpp
```diff
@@ -0,0 +1,268 @@
...
+#include "../headers/config.hpp"
+#include "../headers/comUtil.hpp"
+
+
+/* Module Variable Definitions */
+int commandQueue[COMMAND_QUEUE_SIZE] = {};  // initializes as all zeros
+static int queueIndex = 0;                  // next available index of command queue
+static bool isPaused = false;               // indicates if command execution is paused
+
+// TODO: set Serial.setTimeout() in config.hpp
```
JashanChopra: Is this TODO waiting on another feature?

FSW/src/util/comUtil.cpp
```diff
@@ -0,0 +1,268 @@
...
+ */
+int readCommand()
+{
+    // read Serial
+    if (Serial.available() > 0) // check if there is serial input
+    {
+        int serialInput = Serial.parseInt(); // read the command
+        if ((serialInput <= static_cast<int>(Command::DO_NOTHING)) && (serialInput > 0)) // check if command is valid
+        {
+            //Serial.println("Command Recieved."); // for debugging
```
JashanChopra, line 79: For future debugging or is this leftover from your own testing? Will you be able to change this code on flight?

JashanChopra: Side comment: Will there be a telemetry module? Command received counts are a typical industry telemeter.

FSW/src/util/comUtil.cpp
```diff
@@ -0,0 +1,268 @@
...
+            return serialInput;
+        }
+        else
+        {
+            Serial.println("Invalid command recieved.");
+            Serial.println("(Command Handling)");
```
JashanChopra, line 89: Is there an extra set of parentheses here or is that just for looks?

FSW/src/util/comUtil.cpp
```diff
@@ -0,0 +1,268 @@
...
+        
+        case ENTER_PRE_SUNRISE_MODE: 
+            // TODO: call corresponding function in timing module
+            break;
+        
+        case ENTER_SUNRISE_MODE: 
+            // TODO: call corresponding function in timing module
+            break;
+        
+
```
JashanChopra: extra newline?

FSW/src/util/comUtil.cpp
```diff
@@ -0,0 +1,268 @@
...
+            break;
+        
+        case HEATER_ON: 
+            // TODO: call corresponding function in housekeeping module
+            break;
+        
+        case HEATER_OFF: 
+            // TODO: call corresponding function in houskeeping module
+            break;
+
```
JashanChopra, line 241: ^^

JashanChopra (approved):

FSW/src/util/comUtil.cpp
Vysares: Just for looks, the parentheses get printed

FSW/src/util/comUtil.cpp
Vysares: This is leftover from my own testing, but can be uncommented without consequence. I left it in there just in case we want it for debugging later.

FSW/src/util/comUtil.cpp
Vysares: Good call. Changed to print out the value of the invalid command and the range of acceptable commands.

FSW/src/util/comUtil.cpp
Vysares: Nope, good catch. This will actually go in the initialization function in main. Removed.

FSW/src/util/comUtil.cpp
Vysares: Fixed

FSW/src/util/comUtil.cpp
Vysares: The closest thing to a telemetry module is probably housekeeping. We can easily keep track of received commands since the last restart, but unless we want to store that data on the flash modules there is no way to keep track of it through a restart of the teensy.

PR #11, opened 2021-02-10
Timing module

jaretanderson:
Created a class to track the `ScienceMode` of the payload, corresponding to the format implemented in the command module. Created helper functions to change the payload mode according to the information presented by ADCS and the Photodiode. Also includes a function to determine if the ADCS should reverse its sweep direction, which modifies a sweep event that could be used in an ADCS module developed in the future. 

I had to split the headers for this module into a class header and a function definitions header because there was a circular compile dependency, so let me know if any of the formatting/naming in this module is not clear.

I also added in the proper calls to the `cmdUtil` module to change payload state, but make sure this aligns with what you had intended in those TODOs. 

(also great work with the events classes, using those made this module infinitely easier to develop!!)


FSW/src/headers/config.hpp
```diff
@@ -30,6 +34,11 @@ const int SPI_MAX_SPEED = 2000000; // clock speed of 2MHz for SPI
 const int SERIAL_BAUD = 9600;
 const int SERIAL_TIMEOUT = 50;      // time to wait for serial input, ms
 
+/* - - - - - - ADCS - - - - - - */
+// NanoSAM II will assume that ADCS will be implemented later, but have places in the logic
+// for a future team to slot in these ADCS measurements 
+const bool ADCS_READY_FOR_SCIENCE = true;     // flag on whether or not attitude is ready for science
```
Vysares, line 40: This should be a static bool, so it can be changed

FSW/src/headers/timing.hpp
```diff
@@ -0,0 +1,40 @@
...
+    // TODO: Do we actually need this check? Maybe a bit flip could lead to invalid states being returned? 
+    MODE_NOT_RECOGNIZED        // KEEP AS LAST IN ENUM, used to check that a valid mode is being used
+};
+
+/* - - - - - - Declarations - - - - - - */
+void updatePayloadMode(float buffer[BUFFERSIZE], int bufIdx);
+float smoothBuffer(float buffer[BUFFERSIZE], int bufIdx);
+void checkSweepChange(float buffer[BUFFERSIZE], int bufIdx);
+int wrapBufferIdx(int idx);
+// TODO recoverPayloadMode();
```
Vysares, line 39: Is this for recovering the mode after a restart?

FSW/src/headers/timingClass.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+/* - - - - - - Includes - - - - - - */
+// C++ libraries
+
+// Other libraries
+
+// NS2 headers
+
+/* - - - - - - Class Definitions - - - - - - */
+class ScienceMode
```
Vysares, line 12: Is there any reason to make this a class rather than just a static Mode enum? It seems simpler to do:
    static Mode scienceMode;
    scienceMode = STANDBY_MODE;
instead of:
    static ScienceMode scienceMode();
    scienceMode.set(STANDBY_MODE)

Vysares (approved):
With those changes we talked about this looks good

FSW/src/headers/config.hpp
```diff
@@ -52,10 +61,18 @@ const int FILESIZE = BUFFERSIZE + TIMESTAMP_SIZE;
...
 const int WINDOW_LENGTH_MSEC = WINDOW_LENGTH_SEC * 1000; // milliseconds, length of science data
+const int SWEEP_TIMEOUT_MSEC = 1000; // milliseconds, time for ADCS to sweep optic across the sun
 
 // Events
 static RecurringEvent dataProcessEvent(SAMPLE_PERIOD_MSEC); // assuming that duration arg is ms
 static Event saveBufferEvent;
+static TimedEvent sunriseTimerEvent(WINDOW_LENGTH_MSEC);
+static TimedEvent sweepTimeoutEvent(SWEEP_TIMEOUT_MSEC);
+
+// FUTURE TEAMS: this event is invoked when the ADCS should switch its sweep direction
```
JashanChopra, line 72: This is hot

JashanChopra: Better get a damn thank you letter in the mail

FSW/src/headers/timing.hpp
```diff
@@ -0,0 +1,40 @@
...
+    SAFE_MODE,      // KEEP FIRST IN ENUM, disable all operations
+
+    // add new modes here
+    STANDBY_MODE,   // waiting for wakeup signal from bus
+    SUNSET_MODE,    // gathering data, waiting for phododiode threshold drop
+    PRE_SUNRISE_MODE, // gathering data, waiting for photodiode threshold rise
+    SUNRISE_MODE,   // gathering data on timer (returns to standby)
+
+    // end of list
+    // TODO: Do we actually need this check? Maybe a bit flip could lead to invalid states being returned?
```
JashanChopra, line 30: Was this TODO discussed, I am curious as well.

FSW/src/headers/timingClass.hpp
```diff
@@ -0,0 +1,24 @@
...
+{
+    protected:
+        int mode; // according to mode enum
```
JashanChopra, line 15: Why protect `mode`?

FSW/src/util/comUtil.cpp
```diff
@@ -207,23 +207,28 @@ void executeCommand(int command)
...
         case ENTER_SAFE_MODE: 
-            // TODO: call corresponding function in timing module
+            scienceMode.set(SAFE_MODE);
+            Serial.println("Command Recieved - Entering Safe Mode. This mode must be manually exited.");
             break;
         
         case ENTER_STANDBY_MODE: 
-            // TODO: call corresponding function in timing module
+            scienceMode.set(STANDBY_MODE);
+            Serial.println("Command Recieved - Entering Standby Mode. This mode must be manually exited.");
```
JashanChopra, line 216: Is there an intermediate mode between this standby mode and a data collection mode? If a manual command is set to leave standby mode, what mode is automatically entered. Does it depend on the location in orbit?

FSW/src/util/memUtil.cpp
```diff
@@ -55,6 +58,17 @@ float dataProcessing()
     digitalWrite(PIN_ADC_CS, HIGH);  // set Slave Select pin to high to de-select chip
     SPI.endTransaction();
 
+    /* NOTE TO FUTURE TEAMS:
+     *      Append pointing data to the voltage here
```
JashanChopra, line 62: Pointing Data?

FSW/src/util/timing.cpp
```diff
@@ -0,0 +1,237 @@
...
+    int currentMode = scienceMode.get();
+    
+    if ((currentMode < MODE_NOT_RECOGNIZED) && (currentMode > STANDBY_MODE)){
+        
+        /*FUTURE TEAMS: Flesh out the below if statement with the logic for 
+        *   controlling the payload mode based on ADCS outputs 
+        * For ground testing this is hardcoded as true in config.hpp,
+        *   but once ADCS is implemented you will need to dynamically update this,
+        *   i.e.    ADCS_READY = (digitalRead(BUS_READY_PIN) == HIGH); */
+        if (ADCS_READY_FOR_SCIENCE){
```
JashanChopra, line 69: Variable name is maybe a bit confusing. Is this ADCS as in attitude determination, or ADC's as in multiple analog to digital converters. I don't know if I get the concept of the ADCS being ready for science? I guess the ADCS has to trigger bus reaction wheels to start scanning at the same time science is being collected, so in that case it's valid.

FSW/src/util/timing.cpp
```diff
@@ -0,0 +1,237 @@
...
+        //       do we want some automated way to leave for testing?
+        //       it honestly might be best to leave it as having to manually start
+        //       a data window for testing
+        
+        // execute commands  
+        commandHandling();
+
+        // TODO: memory scrubbing  
+
+    } else { // mode was not recognized, this should not happen
```
JashanChopra, line 136: nice

FSW/src/util/timing.cpp
```diff
@@ -0,0 +1,237 @@
...
+ * 
+ * Outputs:
+ *  smoothVoltage - average of the previous SMOOTH_IDX_COUNT indices of buffer
+ */
+float smoothBuffer(float buffer[], int bufIdx){
+    float smoothVoltage = 0;
+    int i, idx;
+    for (i = bufIdx; i > bufIdx - SMOOTH_IDX_COUNT; i--){ // decrement to move backwards in time
+        
+        idx = i; // I am scared to toss the iterator for this loop through wrapBufferIdx
```
JashanChopra, line 160: what is this c++ malarkey

FSW/src/util/timing.cpp
```diff
@@ -0,0 +1,237 @@
...
+        
+        // ensure that index is within array bounds 
+        // (i.e. starting with i=0 would throw error if we didn't wrap the index)
+        idx = wrapBufferIdx(idx); 
+
+        smoothVoltage += buffer[idx]; // sum voltages
+    }
+    return smoothVoltage / (float)SMOOTH_IDX_COUNT; // return mean
```
JashanChopra, line 168: When I first saw the function name, I was like "It's just gonna be a mean", and it was just a mean. I think I'd prefer a more idiomatic name, like `voltageMean`, or `runningVoltageMean`.

FSW/src/util/timing.cpp
```diff
@@ -0,0 +1,237 @@
...
+            return idx;
+        }
+    }
+    Serial.print("wrapBufferIdx failed to find valid index after ");
+    Serial.print(NUM_TRIES);
+    Serial.println(" tries.");
+    Serial.println("Defaulting to an index of 0 (Timing module)");
+    
+    // TODO: Should this return -1 instead to throw an error?
+    return 0;
```
JashanChopra, line 199: Do you need any return here outside of the if statement? Seems like the if statement covers any possible case? I'd have this return -1 because something is definitely fucked in the event that if block doesn't return a value. If you return zero, you may end up in a case that's quite hard to debug.

JashanChopra (approved):
Couple of nitpicks, either way it's superb.

FSW/src/util/comUtil.cpp
jaretanderson: There is no "leave standby mode" command, only commands to send the payload to one of the other states related to data gathering. `SUNSET_MODE` will wait for the photodiode output to drop to zero then save the data, `SUNRISE_MODE` saves the buffer after a timer, etc.

FSW/src/util/memUtil.cpp
jaretanderson: made this more clear as "attitude data"

FSW/src/headers/timingClass.hpp
jaretanderson: I added more stuff to this class to justify it being a class instead of just an enum, in this commit you looked at it wasn't really justified tbh

FSW/src/headers/timing.hpp
jaretanderson: It has not been. I am just going to keep the check in because it isn't too hard, but I cannot really envision a scenario where the mode is set to an extraneous mode since we only set it with enum variables and are not setting it with manual integers

FSW/src/headers/config.hpp
jaretanderson: hahaha true

FSW/src/util/timing.cpp
jaretanderson: made it more clear as "ADCS_POINTING_AT_SUN", in general in this code ADCS is always attitude determination and control system. I will write some proper documentation on how the ADCS should be implemented with this existing code so that future teams can hit the ground running

FSW/src/util/timing.cpp
jaretanderson: I like `runningVoltageMean` I will make that change

FSW/src/util/timing.cpp
jaretanderson: yeah it shouldn't ever return here, I will change it to -1 so that we definitely know what went wrong

FSW/src/headers/timingClass.hpp
JashanChopra: I'm just wondering why `mode` is a protected class method and not public

PR #12, opened 2021-02-12
timing module cleanup and timestamp append in memUtil

jaretanderson:
Ironed out syntax for including `TimedEvent` object inside `ScienceMode` class, there were some compile order struggles here. This means that I set the value of my `TimedEvent` object inside this class using a `setDuration()` call in a new `init()` function in `main.cpp`. This is because I want to define this time with a const macro, but I have to define that macro in `config.hpp`. This config file must be included after `timingClass.hpp` since config uses `ScienceMode` objects inside it. This creates a cyclical dependency between these two files if I want to instantiate a `TimedEvent` object using a macro inside the `ScienceMode` class. Therefore, I overloaded the constructor for `TimedEvent` to have a default duration of 1 second if no duration is specified at declaration. Then, in `init()` I change the duration to the proper number, defined by the constant macro in `config.hpp`. 

I also laid the groundwork for timestamping the photodiode files and addressed some feedback from the timing module PR, but that is pretty self explanatory.

### DIFF
--- a/FSW/main.cpp
+++ b/FSW/main.cpp
@@ -35,9 +35,13 @@ int main()
     // TODO: Change main loop to account for critical exit conditions
     while(true) // run main loop forever
     {
+        // execute commands  
+        commandHandling();
+
         // test multiple file structure by calling functions from memUtil.cpp
-        scienceMemoryHandling();
-        break;
+        if (scienceMode.get() != SAFE_MODE){
+            scienceMemoryHandling();
+        }
     }
 
     return 0;
--- a/FSW/main.cpp
+++ b/FSW/main.cpp
@@ -66,7 +66,7 @@ int main()
         return -1;
     }
     
-    while(true) // run main loop forever
+    while(true) // run main loop until exit command
     {
         // execute commands  
         commandHandling();
--- a/FSW/main.cpp
+++ b/FSW/main.cpp
@@ -14,9 +14,31 @@
 // Other libraries
 
 // NS2 headers
+#include "src/headers/config.hpp"
+#include "src/headers/comUtil.hpp"
+#include "src/headers/eventUtil.hpp"
+#include "src/headers/timingClass.hpp"
+#include "src/headers/timing.hpp"
 #include "src/headers/memUtil.hpp"
 
 /* - - - - - - Functions - - - - - - */
+
+/* - - - - - - init - - - - - - *
+ * Usage:
+ *  initializes NanoSAM II FSW
+ * 
+ * Inputs:
+ *  none
+ *  
+ * Outputs:
+ *  initStatus - flag (true if initialization was successful)
+ */
+bool init(){
+    scienceMode.setPointingAtSun(true);
+    scienceMode.setMode(SAFE_MODE); // this is done in the class constructor so we may not need it
+
+    return true;
+}
 
 /* - - - - - - main - - - - - - *
  * Usage:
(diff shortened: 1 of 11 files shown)


FSW/main.cpp
```diff
@@ -14,10 +14,40 @@
...
+ *  none
+ *  
+ * Outputs:
+ *  initStatus - flag (true if initialization was successful)
+ */
+bool init(){
+    Serial.print("Initializing NanoSAM II FSW... ");
+
+    scienceMode.setPointingAtSun(true);
+    scienceMode.setMode(SAFE_MODE); // this is done in the class constructor so we may not need it
```
JashanChopra, line 40: Why do we start in safe mode and not the nominal mode?

FSW/main.cpp
```diff
@@ -14,10 +14,40 @@
...
+
+    // lockout duration has to be initialized with a bad value for compilation, overwrite
```
JashanChopra, line 42: Not really understanding this

FSW/src/util/memUtil.cpp
```diff
@@ -214,10 +214,49 @@ bool saveBuffer(int &index){
...
+ *       ([YYYY]-[MM]-[DD]T[hh]:[mm]:[ss].[sss] where [sss] is subseconds)
+ * 
+ * Inputs:
+ *  none
+ *  
+ * Outputs:
+ *  relative timestamp
+ */
+unsigned long calcTimestamp(){
+    unsigned long currentTimeRelative = millis(); // milliseconds
```
JashanChopra, line 253: I'm guessing all this hooplah is just a nicety for future teams?

FSW/src/util/timing.cpp
```diff
@@ -195,8 +211,8 @@ int wrapBufferIdx(int idx){
     Serial.println(" tries.");
     Serial.println("Defaulting to an index of 0 (Timing module)");
     
-    // TODO: Should this return -1 instead to throw an error?
-    return 0;
+    // Return -1 to throw an error since something has gone very wrong if we get here
```
JashanChopra, line 214: So it seems like -1 is being used as the error code in a lot of things. If we see this, will it be obvious where the code fails? Or should these -1 returns be preceded by something like `Serial.println("ouch this hurts")`

JashanChopra (approved):

FSW/src/util/timing.cpp
jaretanderson: this one is preceded by a warning "more than 10 tries, defaulting to buffer idx of 0", so yes I agree

FSW/src/util/memUtil.cpp
jaretanderson: yes hopefully they understand that they need to do something here based on the paragraph of comments

FSW/main.cpp
jaretanderson: should be `STANDBY_MODE`, updated. Good catch

FSW/main.cpp
jaretanderson: Based on the compile order of our headers, there was a cyclic dependency between the location of the sweep lockout macro and the ScienceMode class definition. I circumvented this by initializing `sweepLockoutEvent` with a default timeout (1sec) then overwriting that with the proper macro in `init()`

I will make the comment more descriptive to reflect this

FSW/src/headers/eventUtil.hpp
```diff
@@ -36,6 +36,7 @@ class TimedEvent : public Event
 
     public:
         TimedEvent(unsigned long newDuration);
+        TimedEvent(); // overloaded constructor to set default duration
```
Vysares, line 39: Good call

Vysares (approved):

PR #13, opened 2021-02-14
Housekeeping module

Vysares:
Housekeeping module complete. Housekeeping data at any given time is store in latestHksample, which is a struct containing all the temperature and power data. Previous samples are stored in hkData, which is just an array of many structs. 
Also wrote a MATLAB script for generating the thermistor voltage/temperature lookup table. I'm not sure what realistic temperature range is, the the range of that lookup table and the values of all the min/max temperatures in config.hpp are just guesses for now.

### DIFF
--- a/FSW/src/headers/config.hpp
+++ b/FSW/src/headers/config.hpp
@@ -117,7 +117,7 @@ const float PG_VOLTAGE_MIN_EXPECTED = 3.2;  // volts, minimum expected reading f
 const int HK_SAMPLE_PERIOD_MSEC = 1000;    // milliseconds, interval between housekeeping updates
 
 // Events
-static RecurringEvent housekeepingEvent(HK_PERIOD_MSEC);
+static RecurringEvent housekeepingTimer(HK_SAMPLE_PERIOD_MSEC);
 
 
 /* - - - - - - Timing Module - - - - - - */
--- a/FSW/src/headers/config.hpp
+++ b/FSW/src/headers/config.hpp
@@ -100,14 +100,15 @@ const int WD_PULSE_DUR_MICROSEC = 10;       // microseconds, watchdog reset sign
 const int HK_SAMPLES_TO_KEEP = 100;   // number of previous housekeeping samples to keep
 
 // heater cutoff temperatures
-const float HEATER_TEMP_LOW = 0;     // celsius, heater will turn on at or below this temp
+static bool forceHeaterOn = false;   // if true, heater will always be on regardless of temperature
+const float HEATER_TEMP_LOW = -20;     // celsius, heater will turn on at or below this temp
 const float HEATER_TEMP_HIGH = 20;   // celsius, heater will turn off at or above this temp
 
 // safe temperature range
-const float OPTICS_TEMP_MIN_SAFE = -20; // celsius, minimum safe photodiode temp
-const float OPTICS_TEMP_MAX_SAFE = 50;  // celsius, maximum safe photodiode temp
-const float BOARD_TEMP_MIN_SAFE = 0;    // celsius, minimum safe board temp
-const float BOARD_TEMP_MAX_SAFE = 50;   // celsius, maximum safe board temp
+const float OPTICS_TEMP_MIN_SAFE = -35; // celsius, minimum safe photodiode temp
+const float OPTICS_TEMP_MAX_SAFE = 70;  // celsius, maximum safe photodiode temp
+const float BOARD_TEMP_MIN_SAFE = -20;    // celsius, minimum safe board temp
+const float BOARD_TEMP_MAX_SAFE = 70;   // celsius, maximum safe board temp
 
 // power supply expected voltage range
 const float PG_VOLTAGE_MAX_EXPECTED = 3.4;  // volts, maximum expected reading from "power good" pin
--- a/FSW/src/headers/config.hpp
+++ b/FSW/src/headers/config.hpp
@@ -32,6 +32,7 @@ const int TEENSY_ADC_BINS = 1023;       // bins, number of bins in Teensy ADC
 const float TEENSY_HIGH_VOLTAGE = 3.3;  // volts, max teensy voltage
 const float TEENSY_LOW_VOLTAGE = 0.0;   // volts, min teensy voltage
 const float TEENSY_VOLTAGE_RANGE = TEENSY_HIGH_VOLTAGE - TEENSY_LOW_VOLTAGE; // volts, Teensy ADC voltage range
+const float TEENSY_VOLTAGE_RES = TEENSY_VOLTAGE_RANGE / TEENSY_ADC_BINS;     // volts per bin
 
 /* - - - - - - SPI - - - - - - */
 const int ADC_MAX_SPEED = 2000000; // Hz, maximum SPI clock speed for ADC
@@ -81,11 +82,15 @@ const int WD_PULSE_DUR_MICROSEC = 10;       // microseconds, watchdog reset sign
 /* - - - - - - Housekeeping Module - - - - - - */
 const int HK_SAMPLES_TO_KEEP = 20;   // number of previous housekeeping samples to keep
 
+// heater cutoff temperatures
+const float HEATER_TEMP_LOW = 0;     // celsius, heater will turn on at or below this temp
+const float HEATER_TEMP_HIGH = 20;   // celsius, heater will turn off at or above this temp
+
 // safe temperature range
-const float MIN_SAFE_OPTICS_TEMP = 0;  // celsius, minimum safe photodiode temp
-const float MAX_SAFE_OPTICS_TEMP = 50; // celsius, maximum safe photodiode temp
-const float MIN_SAFE_BOARD_TEMP = 0;   // celsius, minimum safe board temp
-const float MAX_SAFE_BOARD_TEMP = 50;  // celsius, maximum safe board temp
+const float MIN_SAFE_OPTICS_TEMP = -20; // celsius, minimum safe photodiode temp
+const float MAX_SAFE_OPTICS_TEMP = 50;  // celsius, maximum safe photodiode temp
+const float MIN_SAFE_BOARD_TEMP = 0;    // celsius, minimum safe board temp
+const float MAX_SAFE_BOARD_TEMP = 50;   // celsius, maximum safe board temp
 
 // thermistor constants
 const float THERM_VCC = 3.3;           // volts
--- a/FSW/src/headers/config.hpp
+++ b/FSW/src/headers/config.hpp
@@ -27,6 +27,12 @@ const int PIN_ANALOG_THERM = 15;    // analog board thermistor pin
 const int PIN_OPTICS_THERM = 16;    // optics bench thermistor pin
 
 
+/* - - - - - - Teensy ADC - - - - - - */
+const int TEENSY_ADC_BINS = 1023;       // bins, number of bins in Teensy ADC
+const float TEENSY_HIGH_VOLTAGE = 3.3;  // volts, max teensy voltage
+const float TEENSY_LOW_VOLTAGE = 0.0;   // volts, min teensy voltage
+const float TEENSY_VOLTAGE_RANGE = TEENSY_HIGH_VOLTAGE - TEENSY_LOW_VOLTAGE; // volts, Teensy ADC voltage range
+
 /* - - - - - - SPI - - - - - - */
 const int ADC_MAX_SPEED = 2000000; // Hz, maximum SPI clock speed for ADC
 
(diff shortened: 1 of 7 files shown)


FSW/src/headers/config.hpp
```diff
@@ -74,13 +85,40 @@ static TimedEvent sweepTimeoutEvent(SWEEP_TIMEOUT_MSEC);
...
 /* - - - - - - Command Handling Module - - - - - - */
 const int COMMAND_QUEUE_SIZE = 100;     // maximum number of commands the command queue can store.
 const int SERIAL_TIMEOUT_MSEC = 50;     // milliseconds, maximum time to wait for serial input
 
+
 /* - - - - - - Fault Mitigation Module - - - - - - */
-const int WD_RESET_INTERVAL = 100;  // watchdog feeding interval, ms
-const int WD_PULSE_DUR = 10;        // watchdog reset signal duration, MICROSECONDS!!!
+const int WD_RESET_INTERVAL_MSEC = 100;     // milliseconds, watchdog feeding interval
+const int WD_PULSE_DUR_MICROSEC = 10;       // microseconds, watchdog reset signal duration
```
jaretanderson, line 96: I like all of the standardization you did in `config.hpp`, makes it much clearer what the constants mean

FSW/src/headers/hkUtil.hpp
```diff
@@ -0,0 +1,52 @@
...
+    float analogCurrent;   // analog board current, amps
+    float digitalCurrent;  // digital board current, amps
+    float digitalRegPG;    // digital regulator "power good" signal, volts
+};
+
+static HousekeepingData latestHkSample;
+
+struct TempVoltagePair { float temperature, voltage; };
+
+// thermistor temperature/voltage lookup table
```
jaretanderson:
```suggestion
// thermistor temperature (deg C) / voltage (Volts) lookup table
```

FSW/src/util/hkUtil.cpp
```diff
@@ -0,0 +1,184 @@
...
+{
+    if (latestHkSample.opticsTemp >= HEATER_TEMP_HIGH)
+    {
+        digitalWrite(PIN_HEAT, LOW); // turn heater off
+    }
+    else if (latestHkSample.opticsTemp <= HEATER_TEMP_LOW)
+    {
+        digitalWrite(PIN_HEAT, HIGH); // turn heater on
+    }
+}
```
jaretanderson: do you think it is necessary to implement a thermostat algorithm instead of a direct if-statement? 

Right now if the temp is hovering right around `HEATER_TEMP_LOW`/`HIGH` then we could end up turning the heater on and off in a fast oscillating manner every time the loop is called, which could potentially cause electrical noise

Depending on the frequency that `handleHousekeeping()` is called with though (i,e. if it is ~1Hz) then this probably isn't something we need to worry about.

jaretanderson (approved):
Donavon was doing some research on thermal maxes/mins for the thermal test plan, so we should talk to him and get confirmation on the config numbers but the logic here looks good to me

FSW/src/util/hkUtil.cpp
Vysares: If the temp drops below HEATER_TEMP_LOW the heater is turned on and it will stay on until the temperature increases above HEATER_TEMP_HIGH, then it will turn off and stay off until the temp drops back down. That way we get a slow oscillation between the LOW and HIGH temps. As long as the range between them is not tiny this shouldn't trigger too rapidly.

JashanChopra:
> Donavon was doing some research on thermal maxes/mins for the thermal test plan, so we should talk to him and get confirmation on the config numbers but the logic here looks good to me

Donavon said he found -25C to 70C. For this semester, they won't test outside of that range. For actual FSW, you'd probably want to go below that range so that you can check for when the temperature gets too low, and if it isn't corrected quickly, perhaps a transition to safe mode can be applied.

FSW/src/headers/config.hpp
```diff
@@ -69,13 +80,45 @@ static Event saveBufferEvent;
...
+
+
+/* - - - - - - Housekeeping Module - - - - - - */
+const int HK_SAMPLES_TO_KEEP = 100;   // number of previous housekeeping samples to keep
```
JashanChopra: How much time would this cover? I didn't do the math. I imagine we should keep _all_ the housekeeping data between downlink periods for debugging. Data science might want to link temperatures with the data on the ground.

FSW/src/headers/config.hpp
```diff
@@ -69,13 +80,45 @@ static Event saveBufferEvent;
...
+
+// heater cutoff temperatures
+const float HEATER_TEMP_LOW = 0;     // celsius, heater will turn on at or below this temp
```
JashanChopra: I think Danny was saying something about the fact that the heater should basically always be on. Probably still good to have the upper limit though. 

I would design the heater to be always on by default, but I don't really like the idea of that for testing, draws a lot of power.

FSW/src/headers/config.hpp
```diff
@@ -69,13 +80,45 @@ static Event saveBufferEvent;
...
+const float HEATER_TEMP_HIGH = 20;   // celsius, heater will turn off at or above this temp
+
+// safe temperature range
+const float OPTICS_TEMP_MIN_SAFE = -20; // celsius, minimum safe photodiode temp
+const float OPTICS_TEMP_MAX_SAFE = 50;  // celsius, maximum safe photodiode temp
+const float BOARD_TEMP_MIN_SAFE = 0;    // celsius, minimum safe board temp
```
JashanChopra: -20

FSW/src/headers/config.hpp
```diff
@@ -69,13 +80,45 @@ static Event saveBufferEvent;
...
+const float BOARD_TEMP_MAX_SAFE = 50;   // celsius, maximum safe board temp
```
JashanChopra: 70

FSW/src/headers/hkUtil.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+/* - - - - - - Declarations - - - - - - */
+void handleHousekeeping();
+void setHeater();
+void sampleHousekeepingData();
+float voltageToTemp(float voltage);
+void timeSortHkData();
+
+struct HousekeepingData // stores one point of housekeeping data
```
JashanChopra, line 21: Make sure structs like these are documented well outside of the code for reference when future teams (or us later in testing) are looking at the data.

FSW/src/util/hkUtil.cpp
```diff
@@ -0,0 +1,184 @@
...
+// C++ libraries
+
+// Other libraries
+
+// NS2 headers
+#include "../headers/hkUtil.hpp"
+
+
+/* Module Variable Definitions */
+static HousekeepingData hkData[HK_SAMPLES_TO_KEEP] = {}; // array to store housekeeping data
```
JashanChopra: `housekeepingData` for camelCase consistency.

FSW/src/util/hkUtil.cpp
JashanChopra: I don't really see a scenario where the range is exactly on the low or high and it triggers a bunch, but it would probably not be a bad thing to trigger a warning if it hits a bunch, I'm not really sure.

FSW/src/util/hkUtil.cpp
```diff
@@ -0,0 +1,184 @@
...
+    latestHkSample.analogTemp = voltageToTemp(analogTempVoltage);
+    latestHkSample.digitalTemp = voltageToTemp(digitalTempVoltage);
+    latestHkSample.analogCurrent = TEENSY_VOLTAGE_RES*(float)analogRead(PIN_AREG_CURR);
+    latestHkSample.digitalCurrent = TEENSY_VOLTAGE_RES*(float)analogRead(PIN_DREG_CURR);   
+    latestHkSample.digitalRegPG = TEENSY_VOLTAGE_RES*(float)analogRead(PIN_DREG_PG);
+    hkData[hkIndex] = latestHkSample; // update data array
+
+    // Check if values are in acceptable ranges
+    // TODO: Log a fault for every case
+    if (latestHkSample.opticsTemp > OPTICS_TEMP_MAX_SAFE) { } // optics temp too high
```
JashanChopra, line 114: Seems like there's a redundancy between these first two if-else-if statements and the checks in `setHeater()`. I'd either just turn the heater on and off inside `sampleHousekeepingData` or I would put the fault logging in `setHeater()`.

FSW/src/util/hkUtil.cpp
```diff
@@ -0,0 +1,184 @@
...
+    else if (latestHkSample.opticsTemp < OPTICS_TEMP_MIN_SAFE) { } // optics temp too low
+
+    if (latestHkSample.analogTemp > BOARD_TEMP_MAX_SAFE) { } // analog board temp too high
```
JashanChopra, line 117: If any of these are outside of the condition, a transition to safe mode may be validated.

FSW/src/util/hkUtil.cpp
```diff
@@ -0,0 +1,184 @@
...
+ *  based on the data in thermLookup. 
+ *  Will extrapolate beyond max and min entries in thermLookup if needed
+ * 
+ * Inputs:
+ *  voltage - thermistor voltage
+ *  
+ * Outputs:
+ *  temperature - celsius, the temperature corresponding to thermistor voltage reading
+ */
+float voltageToTemp(float voltage)
```
JashanChopra: The table you created, using the MATLAB script (which looks good) is only valid for the analog and digital thermistors. As per these equations: 
image

Recall that the optics thermistor uses a different set of equations: which is recreated via this transfer table: 
image

FSW/src/util/hkUtil.cpp
```diff
@@ -0,0 +1,184 @@
...
+{
+    int thermLookupSize = sizeof(thermLookup) / sizeof(thermLookup[0]); // size of lookup table
+    
+    // find lower adjacent entry
+    int i = 0;
+    while (thermLookup[i].voltage > voltage && i < thermLookupSize - 1) { i++; }
```
JashanChopra, line 147: This looks good. If these calculations do prove to be too intense at the required rate, you can just compare raw voltage values (just take the corresponding voltages at the triggering temperatures). Then we provide good documentation on how to convert voltages to temperatures on the ground for science.

FSW/src/util/hkUtil.cpp
```diff
@@ -0,0 +1,184 @@
...
+ *  Sorts hkData in ascending cronological order and resets hkIndex to 0
+ *  call before using hkData in another function
+ * 
+ * Inputs:
+ *  none
+ *  
+ * Outputs:
+ *  none
+ */
+void timeSortHkData()
```
JashanChopra, line 186: How similar is this to the similar function that sorts the science data by ascending? Can they both be used for the same purpose, or does the `struct` nature of the HK data require this new function?

JashanChopra (changes requested):
Transfer table for the optics thermistor must be implemented. The rest looks solid.

FSW/src/util/hkUtil.cpp
Vysares: There is a difference between the MAX/MIN safe temperatures and the LOW/HIGH heater trigger temperatures. like this:

|MIN_SAFE| <- - - - - |HEATER_LOW(turn on)| <- -(target range)- -> |HEATER_HIGH(turn off)| - - - - - -> |MAX_SAFE|
                                                                            
setHeater checks temperature against HEATER_TEMP_LOW and HEATER_TEMP_HIGH only, and we would expect the temperature to drift a little bit outside that range while the heater is changing state, but if for some reason the temperature drifts well beyond the target range and outside of the safe range we will want to log a fault.

FSW/src/util/hkUtil.cpp
Vysares: If any of these trigger, a fault will be logged, and the fault will generate a command to enter safe mode.

FSW/src/util/hkUtil.cpp
Vysares: Thanks, seems like I missed this. Will implement.

FSW/src/util/hkUtil.cpp
Vysares: Very similar. I had the same thought, but C++ really likes to know exactly what kind of data it is working with, so making generalized timeSortArray function would be more complicated and would likely use more memory than just having the two functions. 
This works for now, but in the future I am thinking of maybe replacing each array with a std::vector. They use more memory, but they are dynamically scalable, and have all kinds of awesome functions like push_back() and erase(), which would allow us to keep the vector sorted all the time.

FSW/src/util/hkUtil.cpp
Vysares: changed.

FSW/src/headers/hkUtil.hpp
Vysares: I will update the readme with info on all the data structures, and more generally to provide an overview of how the software fits together for people who are not me and Jaret.

FSW/src/headers/config.hpp
Vysares: Very good point. It should be somewhere around ~5000 elements large depending on the orbit. With six floats per sample all the housekeeping data will be about 0.5Mbits

FSW/src/headers/config.hpp
Vysares: I added a bool `forceHeaterOn` that `setHeater()` will check first, and if true will just force the heater on all the time.

FSW/src/headers/hkUtil.hpp
Vysares: See new issue

JashanChopra (approved):
Nice work, thanks for the changes.

FSW/src/headers/config.hpp
```diff
@@ -69,13 +80,51 @@ static Event saveBufferEvent;
 static TimedEvent sunriseTimerEvent(WINDOW_LENGTH_MSEC);
 static TimedEvent sweepTimeoutEvent(SWEEP_TIMEOUT_MSEC);
 
+// FUTURE TEAMS: this event is invoked when the ADCS should switch its sweep direction
+//   so link your ADCS module with this event to tell it when to switch direction 
+//      look at checkSweepChange() in timing.cpp for more info
+static Event sweepDirectionChangeEvent; 
+
+
```
jaretanderson, line 88: This event got moved into the `timingClass.hpp` header, it shouldn't be in this file anymore. I wrote a note to circle back and look at the file once you merge it into main, if this shows back up then we will have to remove it, but the merge may take care of that.

jaretanderson (approved):

PR #16, opened 2021-02-18
Refactor

Vysares:
- Renamed everything by module name. Kept eventUtil the same, since it is a project-wide utility, I think that makes sense.
- Enforced consistent braces style, since Jaret and I were using different styles and I didn't like either of them. Opening braces are now on the same line as the preceding statement, with a single space in front. This is the same for classes, functions, switches, conditionals, etc:
```
class SuperCoolClass {
  SuperCoolClass();
}

void superCoolFunction() {
  // do cool stuff
}
```
- `else` and `else if` statements go on the same line as the preceding closing brace. `if` and `switch` statements always have a space between the opening `(` like this:
```
if (something) {
  // do stuff
} else if (somethingElse) {
  // do other stuff
} else {
  // do different stuff
}
```
- Updated `init()` in main.cpp to set all the pinmodes and begin the serial connection. Also added a section to start recurring timers.


jaretanderson (approved):
Good changes, this makes the codebase more cohesive. The new filenames are much more descriptive

JashanChopra (approved):
Not gonna check every line on this one, I trust you to refactor as you see fit. What you laid out seems solid, I'm glad code consistency is a priority as well - it will make it a lot easier for future teams to adopt.

PR #17, opened 2021-02-19
Downlink function and command

jaretanderson:
Wrote the downlink to work asynchronously because I was worried that downlinking many files in a row would take longer than the watchdog timeout. This implementation will downlink one file per main loop until there are no more files available to downlink.

### DIFF
--- a/FSW/src/headers/commandHandling.hpp
+++ b/FSW/src/headers/commandHandling.hpp
@@ -49,7 +49,7 @@ enum Command { // all possible commands
     ADCS_POINTING_AT_SUN_F,         // sets pointing at sun flag to false
 
     // downlink
-    DOWNLINK_ENABLE,                // start downlink at next available time
+    DOWNLINK_START,                // start downlink at next available time
 
     // Main Loop Commands
     EXIT_MAIN_LOOP,
--- a/FSW/src/headers/commandHandling.hpp
+++ b/FSW/src/headers/commandHandling.hpp
@@ -48,6 +48,9 @@ enum Command { // all possible commands
     ADCS_POINTING_AT_SUN_T,         // sets pointing at sun flag to true
     ADCS_POINTING_AT_SUN_F,         // sets pointing at sun flag to false
 
+    // downlink
+    DOWNLINK_ENABLE,                // start downlink at next available time
+
     // Main Loop Commands
     EXIT_MAIN_LOOP,
 
--- a/FSW/src/util/commandHandling.cpp
+++ b/FSW/src/util/commandHandling.cpp
@@ -270,7 +270,7 @@ void executeCommand(int command) {
             break;
 
         // Downlink
-        case DOWNLINK_ENABLE:
+        case DOWNLINK_START:
             scienceMode.downlinkEvent.invoke();
             Serial.println("Command Received - Downlink will begin when payload enters standby.");
             break;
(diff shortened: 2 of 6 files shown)


FSW/src/util/dataCollection.cpp
```diff
@@ -200,7 +206,7 @@ bool saveBuffer(int &index) {
     if (status) {
         Serial.print("Found file ");
```
JashanChopra, line 207: Does this also need the `println` ? Why just change the one below?

FSW/src/util/dataCollection.cpp
```diff
@@ -256,4 +262,61 @@ unsigned long calcTimestamp() {
...
+            Serial.println(":");
+
+            char downlinkBuffer[FILESIZE];
+
+            // read data from file into downlink buffer
+            SerialFlashFile downlinkFile;
+            downlinkFile = SerialFlash.open(filename);
+            if (downlinkFile) {
+                downlinkFile.read(downlinkBuffer, FILESIZE);
+                Serial.println(downlinkBuffer);
```
JashanChopra, line 301: So right now the downlink just prints the data? That's fine, just wanted to make sure I understood this.

JashanChopra (approved):
So I imagine the bus would handle downlink. The question then becomes, do we send over lines of data for downlink, or do we send over files for downlink? I'm not really sure - but I want to say files. Maybe leave a comment for future teams in this code, like you've done previously?

jaretanderson (commented):
Yes for this year we are just printing to serial since we do not have any bus to pass through. Files probably make sense, it really comes down to what connections we have between the bus and payload. Comments to future teams are a good idea, I will update that

FSW/src/util/dataCollection.cpp
jaretanderson: `println` just adds a newline in at the end, I want these three prints all on the same line. Additionally, string concatenation can have issues on arduino so most examples I have seen use this syntax rather than concatenating the strings and using one `println` call.

FSW/src/util/commandHandling.cpp
```diff
@@ -55,25 +55,25 @@ void commandHandling() {
 
 /* - - - - - - readCommand - - - - - - *
  * Usage:
- *  Reads serial input for an incoming command, checks if it has recieved a valid command,
+ *  Reads serial input for an incoming command, checks if it has received a valid command,
```
Vysares, line 58: Nice catch

FSW/src/headers/commandHandling.hpp
```diff
@@ -48,6 +48,9 @@ enum Command { // all possible commands
     ADCS_POINTING_AT_SUN_T,         // sets pointing at sun flag to true
     ADCS_POINTING_AT_SUN_F,         // sets pointing at sun flag to false
 
+    // downlink
+    DOWNLINK_ENABLE,                // start downlink at next available time
```
Vysares: It's probably a good idea to have a `DOWLINK_DISABLE` command as well

FSW/src/util/dataCollection.cpp
```diff
@@ -256,4 +262,61 @@ unsigned long calcTimestamp() {
...
+void downlink() {
+    
+    if (newDownlink) { // reset asynchronous variables if it is a new downlink
+        filename[FILE_IDX_OFFSET] = '0';
+        downlinkFileIdx = 0;
+        downlinkFileCount = 0;
+        newDownlink = false;
+    }
+
+    if (downlinkFileIdx < MAXFILES){ // check all science possible filenames
```
Vysares, line 287: I really like that you've staggered the downlink over consecutive loop runthroughs.

Vysares (approved):

PR #18, opened 2021-03-01
Scrubbing dev

Vysares:
There's a lot to unpack here. I've made changes to **every single file** and added four more files on top of that. I've devoted so much time to this branch that I am literally dreaming about pointers. Here are the highlights:
- Added `HammingBlock` class, which holds an 8 byte message inside a 9 byte array of encoded data. Each `HammingBlock` object is capable of encoding/decoding itself, and can also scan itself for errors and correct them. I don't expect you guys to understand how this works, but I tested it and it does and it's honestly amazing. I can walk you through it if you want.
- Added `EncodedFile` class, which holds an array of `HammingBlock` objects. An `EncodedFile` can be passed buffer and timestamp data, which it will break up into hamming blocks and encode. Calling `getData()` will return a pointer to an array of contiguous encoded data that can be written to a file. To avoid memory leaks I religiously avoided dynamic allocation of memory, so everything is pre-allocated on the stack.
- Added `AsyncEvent` class, which can be used when we want to stagger an event over multiple loop iterations. 
- Refactored the `downlink()` function to use an `AsyncEvent` and moved the file counter inside the function as a local static variable, that way we don't have variables hanging around at the top of the cpp file that are only used by one function.
- Added `scrubFlash()` function, which uses the same framework as `downlink()`, and scrubs each file in memory, overwriting corrupted files with corrected ones. Also added a command to start a new scrub.
- Moved the logic that determines what to do in each mode to the main loop. This way `updatePayloadMode` is only responsible is determining the mode, and most of the high level logic is easily readable. Also moved `downlinkEvent` and `exitMainLoopEvent` out of `ScienceMode` class and into config for the same reason. @jaretanderson, let me know if those events belonged there for a super good reason and I just broke everything.
- Changed `timeSortHkData()` to `getHkData()` which now returns a pointer to a new sorted array of housekeeping data and doesn't actually effect the housekeeping buffer. 
- Enforced consistent header inclusion, so that header files only include C++ libraries and `config.hpp`, and all other NS2 headers are included in cpp files. This way a header file only grants access to its own members and doesn't drag other headers with it.
- Numerous readability improvements


FSW/src/headers/eventUtil.hpp
```diff
@@ -35,15 +50,58 @@ class TimedEvent : public Event {
...
+ *      If the time has elapsed, checkInvoked() will return true and automatically restart itself.
+ */
 class RecurringEvent : public TimedEvent {
     public:
         RecurringEvent(unsigned long newDuration);
+
+        bool checkInvoked();
+};
+
+/* - - - - - - AsyncEvent - - - - - - *
```
JashanChopra, line 74: This is really exceptional

FSW/src/headers/hammingBlock.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <vector>
+#include <memory>
+
+// Other libraries
+
+// NS2 config and utility headers
+#include "config.hpp"
+
+/* - - - - - - Bit Operation Macros - - - - - - */
+// a = target variable, bit = bit number to act on (0 indexed)
```
JashanChopra: I think the use of the variable name `a` is a bit confusing, but the comment probably allows it

FSW/src/util/commandHandling.cpp
```diff
@@ -269,15 +267,20 @@ void executeCommand(int command) {
...
+        // Memory
         case DOWNLINK_START:
-            scienceMode.downlinkEvent.invoke();
+            downlinkEvent.invoke();
             Serial.println("Command Received - Downlink will begin when payload enters standby.");
             break;
 
+        case SCRUB_FLASH:
+            scrubEvent.invoke();
+            Serial.println("Command Received - Scrub initiated.");
```
JashanChopra: Maybe a touch more explicit: "Memory Scrub Initiated."

FSW/src/util/dataCollection.cpp
```diff
@@ -164,20 +157,26 @@ void updateBuffer(float sample, int &index) {
...
-        timeSortArray[j] = dataBuffer[i];
+        timeSortBuffer[j] = dataBuffer[i];
         j++;
     }
+    
+    // compute timestamp
+    unsigned long timestamp = calcTimestamp(); 
+    
+    // Run all the data through EDAC
+    EncodedFile encodedFileData = EncodedFile(dataBuffer, timestamp); // this one line cost 50+ hours of my life
```
JashanChopra, line 179: I feel this on a spiritual level

FSW/src/util/dataCollection.cpp
```diff
@@ -276,47 +268,108 @@ unsigned long calcTimestamp() {
...
+    static char scrubFilename[] = "scienceFile0.csv"; 
+    static ScrubReport totalScrubInfo;
+
+    EncodedFile correctedFileData;
+    ScrubReport scrubInfo;
+    
+    // reset static variables at start of new event
+    if (scrubEvent.first()) {
+        totalScrubInfo.numErrors = 0;
+        totalScrubInfo.corrected =0;
```
JashanChopra: `= 0`

FSW/src/util/dataCollection.cpp
```diff
@@ -276,47 +268,108 @@ unsigned long calcTimestamp() {
...
+        SerialFlashFile file;
+        file = SerialFlash.open(scrubFilename);
+        if (file) {
+            uint8_t fileContents[ENCODED_FILE_SIZE];
+            file.read(fileContents, ENCODED_FILE_SIZE);
+            correctedFileData.fill(fileContents);
+            scrubInfo = correctedFileData.scrub(); // scrub it
+
+            // update total scrub info
+            totalScrubInfo.corrected += scrubInfo.corrected;
```
JashanChopra, line 349: Is there some check you can do to make sure the data is scrubbed before iterating the variables?

FSW/src/util/encodedFile.cpp
```diff
@@ -0,0 +1,207 @@
...
+ */
+ScrubReport EncodedFile::scrub() {
+    ScrubReport scrubInfo;
+
+    /* scan and correct each block */
+    for (int blockNum = 0; blockNum < MESSAGE_COUNT; blockNum++) { // for each block...
+        // correct the block and get an error report
+        ErrorReport errorInfo = m_blocks[blockNum].correctBlock();
+
+        // if uncorrectable error detected, clear the block
```
JashanChopra, line 116: What I've seen typically in data science is that faulty data is flagged rather than outright deleted. However, I think when it comes to an error like multiple bit flips - there's no point not deleting it, because there's nothing you could do to fix that data correct? Deleting it also has advantages in terms of computing time and saving memory, so I think this is still the better solution, just curious if you had thought of the alternative.

FSW/main.cpp
```diff
@@ -83,18 +84,44 @@ int main() {
...
+            }
+            // downlink data
+            if (downlinkEvent.checkInvoked()) {
+                downlink;
+            }
+        }
+
+        if (exitMainLoopEvent.checkInvoked()) {
+            Serial.print("Exiting main loop at iteration ");
+            Serial.println(mainLoopIterations);
```
jaretanderson, line 124: moving these into `config` instead of `ScienceMode` makes sense, I didn't have them there for a real reason reflecting on it.

FSW/src/headers/eventUtil.hpp
jaretanderson: agreed with jashan, this is so clean

FSW/src/headers/eventUtil.hpp
```diff
@@ -35,15 +50,58 @@ class TimedEvent : public Event {
...
+ * Use an AsyncEvent when you want to stagger a task over multiple loop iterations.
+ *  -To invoke the AsyncEvent call invoke()
+ *  -To change the maximum number of iterations call setMaxIter(). 
+ *  -To get the current iteration call iter().
+ *  -To de-invoke the AsyncEvent and reset the iteration count call stop().
+ *  -To de-invoke the AsyncEvent without resetting the iteration count, call pause().
+ *  -To check if the AsyncEvent is invoked call checkInvoked().
+ *      once the AsyncEvent is invoked, checkInvoked() will return true and increment iteration
+ *      until max iterations is reached or stop() is called.
+ *      If max iterations is not specified, the AsyncEvent will stay invoked until stop() is called.
```
jaretanderson, line 86: good explanation of usage, this is clear

FSW/src/util/eventUtil.cpp
```diff
@@ -8,96 +8,141 @@
...
+
+// returns whether the current iteration is the first
+bool AsyncEvent::first() {
+    return (iteration == 1);
+}
+
+// returns whether the current iteration is the last
+bool AsyncEvent::over() {
+    return (iteration == maxIterations);
+}
```
jaretanderson, line 148: would it perhaps make sense for this to be `iteration >= maxIterations` just in case the over check is skipped but the event is iterated? There is an argument for keeping it the way it is since that likely would point to a logic error in the code anyway.

FSW/src/util/hammingBlock.cpp
```diff
@@ -0,0 +1,269 @@
+/* hammingBlock.cpp defines the HammingBlock class
```
jaretanderson, line 1: im gonna have to mostly trust you on this logic, we can talk through it in the meeting if there are any particular sections you're worried about though I would be happy to review them in detail

jaretanderson (approved):
I feel much better about this since you mentioned you were testing it out. Great work on this, lots of the cleanup changes makes the FSW more readable. Additionally, the async event will be super useful moving forward if runtime seems to be a problem at all

FSW/src/headers/hammingBlock.hpp
Vysares: changed `a` to `trgt`

FSW/src/util/commandHandling.cpp
Vysares: Changed to "Flash scrub initiated."

FSW/src/util/dataCollection.cpp
Vysares: Fixed.

FSW/src/util/encodedFile.cpp
Vysares: Correct, for a multiple bit error, there is no way to know where the error is, so the whole block is unfixable. I guess we could theoretically implement some super cool system that looks at surrounding data points, interpolates the likely range of values for the block, and determines which bits were likely flipped... that would actually be awesome, but no. When we clear a block like this, it will look like four consecutive measurements of zero irradiance, so right now there is no way to tell the difference between a corrupted block and an uncorrupted block that happens to be all zeros. The issue is we can't flag a block without adding extra bits to the file, since the blocks are already at maximum data density, and since we don't know how many blocks are going to be corrupted in any given file, we would need to append a bunch of redundant bits just to store flags. One alternative I had in mind is to overwrite corrupted blocks with a 64 bit int equal to the index of the block in the file. Then when we decode a file, if any of the blocks contain their own index then we can be fairly confident that block was wiped. The likelihood that a real measurement happens to be that same value is extremely low. Of course, any data we write in the file is also vulnerable to errors, so there is no way to win 100% of the time.

FSW/src/util/eventUtil.cpp
Vysares: The only method that iterates `iteration` is `checkInvoked()`, which will automatically reset the event if
`iteration >= maxIterations`, so there is absolutely no way this scenario can occur.

FSW/src/util/dataCollection.cpp
Vysares: `scrubInfo` will default to all zeros if the file didn't find any errors, so `totalScrubInfo` will only change when errors are detected.

FSW/src/util/encodedFile.cpp
JashanChopra: Sounds good, glad to see it's on the mind.

FSW/src/util/hammingBlock.cpp
```diff
@@ -0,0 +1,269 @@
...
+ * Usage:
+ *  A HammingBlock object is a container for an 8 byte message, which is encoded into a 9 byte
+ *  block using a (72,64) Hamming code.
+ *  A HammingBlock can encode/decode itself, scan its contetns for errors, and correct single bit errors.
```
JashanChopra: `contents`

FSW/src/util/hammingBlock.cpp
```diff
@@ -0,0 +1,269 @@
...
+ * Usage:
+ *  Encodes an 8 byte message into the block
+ * 
+ * Inputs:
+ *  message - pointer to a single 8 byte message block, typecast to void*
+ *  
+ * Outputs:
+ *  None
+ */
+void HammingBlock::encodeMessage(void *message) {
```
JashanChopra, line 41: More straightforward than I would have expected. Going to trust you on the algorithm, but this is a perfect example case for where unit testing really helps out, would go nicely in one of the final reports.

FSW/src/util/hammingBlock.cpp
```diff
@@ -0,0 +1,269 @@
...
+        if (indexParity == 0) {
+            // no error detected
+        } else {
+            // two bit error detected
+            errorInfo.size = 2;
+        }
+    } else {
+        // single bit error detected
+        errorInfo.size = 1;
+        errorInfo.position = indexParity; // The parity tells us the index of the error!!!
```
JashanChopra, line 137: Comments are generally doing a solid job explaining what's going on here, some more in detail documentation will definitely be needed here though if we want to make this accessible to next year's team.

JashanChopra (approved):
The nuance of the pointers is lost on me unfortunately, but overall I think the structure and logic with this new block looks great. I'm happy with the responses on the questions, it's clear you've put in the work and got a great grasp of how this works. Unit tests and documentation that are a WIP will go a long way for this submodule. Excellent work!

FSW/src/util/hammingBlock.cpp
Vysares: Yeah, Hamming codes are super elegant to implement thanks to bitwise operations.
Most of the magic happens in `indexParity ^= hammingIdx`

FSW/src/util/hammingBlock.cpp
Vysares: Oh yes, some serious documentation is called for to make this understandable to future teams, especially since they will need to modify it to add pointing data. You should have seen it before I encapsulated this in classes — it worked, but trying to understand it was hopeless

FSW/src/util/hammingBlock.cpp
Vysares: Fixed.

PR #19, opened 2021-03-05
Scrubbing update

Vysares:
- Fixed a bug in `HammingBlock` class that was causing false detections of double bit errors and led to correctable blocks being cleared. 
- Modified `EncodedFile` class, so that calling `getData()` interleaves every encoded block on the individual-bit level. Now burst errors that affect several consecutive bits will cause correctable single bit errors in several blocks, as opposed to just nuking one or two blocks.
- pushed super early WIP of testing GUI, just in case my local storage combusts without warning.


JashanChopra (approved):
Great catch, glad we caught that early. Overwriting good data is definitely the primary concern with this stuff - so it's good we keep that in mind. 

Once again, GUI looks cool. It's also nice to see code that I actually understand haha, if only FSW could be python :).

FSW/src/util/encodedFile.cpp
```diff
@@ -138,8 +149,13 @@ ScrubReport EncodedFile::scrub() {
 uint8_t *EncodedFile::getData() {
     memset(m_data, 0, ENCODED_FILE_SIZE); // clear the data array
     for (int blockNum = 0; blockNum < MESSAGE_COUNT; blockNum++) { // for each block...
-        // copy the encoded block contents to the data array
-        memcpy(m_data + blockNum * HAMMING_BLOCK_SIZE, m_blocks[blockNum].getBlock(), HAMMING_BLOCK_SIZE);
+        // Interlace each block, so that burst errors span several blocks
+        for (int blockBit = 0; blockBit < HAMMING_BLOCK_SIZE * 8; blockBit++) {
```
jaretanderson, line 153: this is a clever solution to maximize the utility of single bit correction!

jaretanderson (approved):

PR #20, opened 2021-03-09
Unit test

jaretanderson:
Super basic unit tests for non-void functions in command, housekeeping, and timing modules. Want to merge this in so that I can start a fresh branch for unit testing the `EncodedSciData` class.


UnitTest/TestScripts/commandTest.cpp
```diff
@@ -0,0 +1,79 @@
...
+            Serial.print("] = ");
+            Serial.print(key[i]);
+            Serial.print("  -  checkMetaCommand(");
+            Serial.print(i);
+            Serial.print(") = ");
+            Serial.println(checkMetaCommand(i));
+        }
+        
+        // call function
+        if (key[i] != checkMetaCommand(i)) {
```
JashanChopra: Seems a little backwards to me to check if the test has failed. Typically you'd check if a test passed:

`if (key[i] == checkMetaCommand(i))` then pass.

I think it's fine to do it the other way, just make sure its consistently checking for test failure in the future.

JashanChopra: I suppose in this case it does reduce the amount of logic required, since you want to print the name of failed tests, so it's probably fine.

UnitTest/TestScripts/commandTest.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * tests checkMetaCommand with all possible commands
+ * 
+ * Inputs:
+ *  none
+ *  
+ * Outputs:
+ *  0 - pass
+ *  1 - fail
+ */
+int testCheckMetaCommand() {
```
JashanChopra: Is this testing `executeCommand` instead of `checkMetaCommand`, I don't see how this test would pass?

JashanChopra (approved):
I would create some test naming convention, so that you know what file the tested functions are coming from. For example, if you are testing functions in `commandHandling.cpp` the test should probably be named `commandHandlingTest.cpp` instead of just `commandTest.cpp`

jaretanderson:
> I would create some test naming convention, so that you know what file the tested functions are coming from. For example, if you are testing functions in `commandHandling.cpp` the test should probably be named `commandHandlingTest.cpp` instead of just `commandTest.cpp`

Agreed, updated

UnitTest/TestScripts/timingTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+    for (int i = 0; i<5; i++) {
+        if (key[i] != voltageRunningMean(testBuf, testIdx[i])){
+            Serial.println("voltageRunningMean() unit test failed (timing module)");
+            return 1;
+        }
+        
+        // double values in buffer for next test
+        for (int j = 0; j < BUFFERSIZE; j++) {
+            testBuf[j] = testBuf[j] * 2;
+        }
```
Vysares, line 79: Wouldn't changing the buffer also change the values of `key`?

UnitTest/TestScripts/timingTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    // build test cases and answers
+    float key[] = {0.200F, 0, 4.000F, 0, 16.000F};
+    int testIdx[] = {0, BUFFERSIZE-2*SMOOTH_IDX_COUNT, 2*SMOOTH_IDX_COUNT, 
+                      BUFFERSIZE-2*SMOOTH_IDX_COUNT, 2*SMOOTH_IDX_COUNT};
+    float testBuf[BUFFERSIZE];
+
+    for (int i = 0; i<BUFFERSIZE; i++) {
+        if (i < BUFFERSIZE/2) {
+            testBuf[i] = 1.000f; // set even idx to 2
```
Vysares, line 63: This just sets the first half of the buffer to 1, to set even indexes you would need to check `if (i % 2 == 0)`

UnitTest/TestScripts/timingTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+ * Outputs:
+ *  0 - pass
+ *  1 - fail
+ */
+int testVoltageRunningMean() {
+
+    // build test cases and answers
+    float key[] = {0.200F, 0, 4.000F, 0, 16.000F};
+    int testIdx[] = {0, BUFFERSIZE-2*SMOOTH_IDX_COUNT, 2*SMOOTH_IDX_COUNT, 
+                      BUFFERSIZE-2*SMOOTH_IDX_COUNT, 2*SMOOTH_IDX_COUNT};
```
Vysares, line 58: why would the key have different values for the same values of `testIdx`? I'm not sure I'm understanding the logic here.

UnitTest/instructions.md
```diff
@@ -0,0 +1,6 @@
+## Unit Test Instructions
+All software compiled on the teensy has to be nested in a `src` directory, so that makes this file structure take a form that may not be intuitive. 
+To run a test on the teensy, first make sure your `FSW/src` directory is up to date with the latest version of the FSW that you want to test. 
+Then, copy the contents of `UnitTest/TestScripts` into `FSW/src`
+This directory in the arduino sketch will now contain at least three folders, `FSW/src/headers`, `FSW/src/util`, and now `FSW/src/TestScripts`
+Once this is complete, you can open `unitTestDriver.c` as your main script in the Arduino IDE and compile and run as usual.
```
Vysares, line 6: This is really nice to have

Vysares (approved):
Looks good, I'm not sure if I understand what's going on with the timing tests, but I suspect that may have more to do with my brain than your logic. See comments. 
I love how all the tests can be run from a main script, it's really clean.

UnitTest/TestScripts/timingTest.cpp
jaretanderson: I was intending to set the first half to 1, that comment is incorrect so I will fix it.

UnitTest/TestScripts/timingTest.cpp
jaretanderson: I hardcoded `key` to be the expected value, so as the `testBuf` doubles inside the function that has been accounted for already with the values that I hardcoded into `key`

UnitTest/TestScripts/timingTest.cpp
jaretanderson: I am doubling the buffer inside the test as mentioned in the above comment, so that is what leads to the key change

PR #22, opened 2021-03-18
Fault mitigation dev

Vysares:
At last, the long overdue fault mitigation module has arrived. Here's the breakdown:
- Added `faultManager.cpp`, which handles two different structs of persistent data: `payloadData` and an array of `faultReport` structs. All persistent data is saved to the Teensy's emulated EEPROM, which is on the program flash.
- `payloadData` holds useful data about the NS2 system like the number of total restarts and number of consecutive unexpected restarts.
- Since there is nowhere near enough room on the EEPROM (only 1080 bytes) to store a complete log of faults, each fault type has one `faultReport`, which records the time and mode of its last occurence, and a count of total occurences. The occurence counter is only one byte, so it maxes out at 255, but if there are 255 of the same fault, we can be pretty sure that it is a recurring problem. I kept it small so we avoid too many writes to the EEPROM each time a fault is logged, it is only reliable up to 100,000 erase/write cycles. I have some ideas about optimizing this, so I will likely go back in and make changes.
- Correction commands are automatically executed for each logged fault. To avoid uselessly executing the same command 255 times for a recurring fault, corrective commands are only executed on the first and 255th occurrence of a fault. The occurrence counter for each fault is reset to 0 when the payload enters standby, but the timestamp and mode of the last occurrence are still stored.
- Now `init()` checks to see if the last restart was expected, and if it wasn't the software enters safe mode.
- `EncodedFile` class is now a class template `EncodedFile<size_t N>`, so it can hold ANY amount of data, as long as the data size `N` is known at compile time. This makes it possible to encode persistent data before writing it to the EEPROM. 
- Made a new non-template class `EncodedSciData`, derived from a special case of `EncodedFile` with extra methods for manipulating the science data.
- Cleaned up the global namespace by moving purely-derived constants into class declarations and wrapping command codes in their own namespace.
- Numerous minor changes like always


FSW/src/headers/encodedFile.hpp
```diff
@@ -27,40 +23,185 @@ struct ScrubReport {
...
+template <size_t N>
+void EncodedFile<N>::encodeData(void *src) {
+    memset(m_data, 0, MEMSIZE); // clear the data array
+   // memcpy(m_data, src, DECODED_MEMSIZE); // copy buffer to data array
+
+    /* encode the data in blocks */ 
+    for (int blockNum = 0; blockNum < MESSAGE_COUNT; blockNum++) { // for each block...
+        // copy a chunk of unencoded data into a temporary message block
+        uint8_t message[HammingBlock::MSG_SIZE]; 
+        memcpy(message, static_cast<uint8_t*>(src) + blockNum * HammingBlock::MSG_SIZE, HammingBlock::MSG_SIZE);
```
jaretanderson, line 98: show this line to someone if you want to scare them away from C++ lol

FSW/src/headers/config.hpp
```diff
@@ -114,13 +110,13 @@ const int WD_PULSE_DUR_MICROSEC = 10;       // microseconds, watchdog reset sign
 const int HK_SAMPLES_TO_KEEP = 5000;   // number of previous housekeeping samples to keep
 
 // heater cutoff temperatures
-static bool FORCE_HEATER_ON = false;   // if true, heater will always be on regardless of temperature
+static volatile bool FORCE_HEATER_ON = false;   // if true, heater will always be on regardless of temperature
```
jaretanderson, line 113: do all statics have to either be declared `const` or `volatile`? What is the behavior of a static with neither of these qualifiers?

FSW/src/headers/encodedFile.hpp
```diff
@@ -27,40 +23,185 @@ struct ScrubReport {
...
+        ErrorReport errorInfo = m_blocks[blockNum].correctBlock();
+
+        // if uncorrectable error detected, clear the block
+        if (errorInfo.size >= 2) {
+            m_blocks[blockNum].clear();
+        } 
+        // update scrub report with latest error report
+        scrubInfo.numErrors += !!errorInfo.size;
+        scrubInfo.corrected += errorInfo.size == 1;
+        scrubInfo.uncorrected += errorInfo.size > 1;
```
jaretanderson, line 170: this is a great solution

FSW/src/headers/encodedSciData.hpp
```diff
@@ -0,0 +1,33 @@
...
+class EncodedSciData : public EncodedFile<SCIDATA_RAW_MEMSIZE> {
+    private:
+        // member variables
+        uint16_t m_buffer[BUFFERSIZE];        // array to hold decoded buffer
+        unsigned long m_timestamp = MEMSIZE;  // file timestamp
+
+    public:
+        // constructors
+        EncodedSciData() { }
+        EncodedSciData(uint16_t *buffer, unsigned long &timestamp);
```
jaretanderson, line 25: what's the point of overloading here? when would we use the default constructor?

FSW/src/util/faultManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    // update corresponding fault report
+    faultLog[code].startNum = payloadData.startCount;
+    faultLog[code].timestamp = millis();
+
+    if (faultLog[code].occurences < 255) {
+        faultLog[code].occurences++;
+        
+        // mark for action on first and max occurences
+        if (faultLog[code].occurences == 1 || faultLog[code].occurences == 255) {
```
jaretanderson, line 66: the reason for checking on the first occurrence is obvious, but why also the max?

FSW/src/headers/faultManager.hpp
```diff
@@ -0,0 +1,67 @@
...
+        // Power
+        DREG_OUT_OF_RANGE,
+        
+        ERR_CODE
+    };
+    const int COUNT = ERR_CODE + 1;
+};
+
+struct FaultReport {
+    uint8_t occurences = 0;
```
jaretanderson, line 37: I believe it is spelled `occurrence` not `occurence`

FSW/src/util/faultManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+            /* Take action to correct fault */
+            // TODO: Expand to include more faults
+            switch (code) {
+                case faultCode::UNEXPECTED_RESTART:
+                    executeCommand(commandCode::ENTER_SAFE_MODE);
+                    break;
+                
+                // temp too hot
+                case faultCode::ANALOG_TOO_HOT:
+                case faultCode::DIGITAL_TOO_HOT:
```
jaretanderson, line 116: are these TODO or are we just not gonna do anything if it is too hot

FSW/src/util/faultManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+ *  Does not stress the EEPROM, since only changed bytes are overwritten. 
+ *  
+ * Inputs:
+ *  None
+ * Outputs:
+ *  None
+ */
+void saveEEPROM() {
+
+    payloadData.eepromWriteCount++; // one more write to EEPROM
```
jaretanderson, line 215: really good thing to have for end of life analysis

jaretanderson (approved):
I really like the way you compress the faults so that they can fit on the limited EEPROM space

FSW/src/headers/config.hpp
Vysares: declaring them as `volatile` makes the compiler stop yelling at us that they are unused, but since you discovered that all the static variables in `config` are useless, this is probably irrelevant and will be changed.

FSW/src/headers/encodedSciData.hpp
Vysares: The default constructor is used when we want to fill the `encodedFile` with data that is already encoded.  An example is memory scrubbing; we construct and empty `encodedFile` with the default constructor and then call `fill(src)` which fills the object with data from `src` so it can be scrubbed.

FSW/src/util/faultManager.cpp
Vysares: The switch statement carries through until a `break`, so these all share the same logic

FSW/src/util/faultManager.cpp
Vysares: If taking corrective action on the first occurrence doesn't resolve the issue, it will try one more time to correct it before the fault report fills up, it's definitely redundant, but I don't think there's any harm in a little redundancy.

jaretanderson: agreed, that makes sense

FSW/src/headers/encodedSciData.hpp
jaretanderson: gotcha, thanks for the explanation

FSW/main.cpp
```diff
@@ -104,8 +111,14 @@ int main() {
...
+            handleFaults();
         }
 
+         /* - ONLY EXECUTE ON ENTRY TO STANDBY MODE - */
+         if (scienceMode.onStandbyEntry.checkInvoked()) {
+            // reset fault occurence counts
+            resetFaultCounts();
+         }
 
          /* - ONLY EXECUTE IN STANDBY MODE - */
```
JashanChopra, line 123: I want to say faults should be handled in standby mode as well.

FSW/src/headers/config.hpp
```diff
@@ -106,6 +97,11 @@ const int COMMAND_QUEUE_SIZE = 100;     // maximum number of commands the comman
 
 
 /* - - - - - - Fault Mitigation Module - - - - - - */
+static volatile bool ACT_ON_NEW_FAULTS = true;
+const int PERSIST_DATA_ADDR = 0; // first address of persistent system data in EEPROM
+const uint8_t EXPECTING_RESTART_FLAG = 0xaa; // 10101010, value of flag indicating that the last restart was expected.
```
JashanChopra, line 102: I'm guessing capitalization doesn't matter on this hex value?

FSW/src/headers/encodedFile.hpp
JashanChopra: =(

FSW/src/util/faultManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+ * 
+ * Outputs:
+ *  None
+ */
+void logFault(uint8_t code) {
+    detectedFault = true;
+
+    // check if fault is valid
+    if (code >= faultCode::ERR_CODE) {
+        Serial.println("Warning, tried to log a fault with invalid fault code (Fault Manager)");
```
JashanChopra, line 51: Perhaps print the invalid fault code with this, to identify what the invalid fault logged was.

FSW/src/util/faultManager.cpp
JashanChopra: I think it would make more sense to try and take corrective action a few more times, then switch the spacecraft to safe mode. If software _should_ be able to correct a fault, and it fails something like 5 times in a row - something else is definitely wrong, and the ground controller should step in.

FSW/src/util/faultManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+ * Usage:
+ *  resets the watchdog
+ *  
+ * Inputs:
+ *  None
+ * Outputs:
+ *  None
+ */
+void feedWD() {
+    // TODO: some checks here to make sure everything is okay before feeding the dog?
```
JashanChopra, line 82: I can't think of any. I suppose you could check that the output of the voltage regulator is good, but if the voltage regulator stops outputting then the Teensy wouldn't be powered anyway. You could check the temperatures, and not send the watchdog signal if the temperature goes too high or low, but then the issue is that if the Teensy doesn't restart and start sending the signal again you can't ever turn it back on. 

I think some more work needs to be done to think of a better watchdog solution, so I wouldn't worry about it for now.

FSW/src/util/faultManager.cpp
```diff
@@ -0,0 +1,303 @@
...
+    // encode the data
+    EncodedFile<EEPROM_DECODED_SIZE> encodedData = EncodedFile<EEPROM_DECODED_SIZE>();
+    encodedData.encodeData(rawData);
+
+    // write to EEPROM
+    for (int byteNum = 0; byteNum < encodedData.MEMSIZE; byteNum++) {
+        int eepromAddress = PERSIST_DATA_ADDR + byteNum;
+        uint8_t dataToWrite = encodedData.getData()[byteNum];
+
+        if (EEPROM.read(eepromAddress) != dataToWrite) { // only overwrite if the byte has changed. This extends the life of the EEPROM.
```
JashanChopra, line 241: neat!

JashanChopra (approved):

FSW/main.cpp
Vysares: The faults are still handled in standby mode, it's just that the fault counters are reset when standby mode is entered.  Since the software passes sequentially through `standby -> sunset -> presunrise-> sunrise -> standby` once per orbit, faults will be tallied over each orbit so systemic issues will stand out. You can think of it as the beginning of a new fault collection cycle.

FSW/src/headers/config.hpp
Vysares: Yep

FSW/src/util/faultManager.cpp
Vysares: Done

FSW/src/util/faultManager.cpp
Vysares: I agree, I'll leave it like that for now. The worst case scenario is still that the software gets caught up indefinitely, which is the point of having a watchdog at all.

FSW/src/util/faultManager.cpp
Vysares: Oooo I like that a lot. I'll save that for a future PR, once we work out some of the inevitable bugs.

PR #23, opened 2021-03-22
Bugfixes and more

Vysares:
Fixed the issue where static objects in config were not referencing properly. Declaring them as static didn't do what we thought, and every file ended up with their own private copies of all variables in `config.hpp`, static or not. 
Now all non-constant global variables are declared as `extern` instead of `static`, which basically tells to compiler "Trust me, this variable is defined somewhere, go look for it." Fun fact: this is how function calls in multiple files can reference the same function definition, the compiler actually interprets `void foo()` as `extern void foo()`.

The actual definitions for each non-constant variable are in a new file, `externalDefinitions.cpp`. The compiler will point every file to the the instances defined here. I added new constants to `config.hpp` for the initialization values of `bool` and `float` variables, so we can still configure everything from `config.hpp`, and `externalDefnitions.cpp` can chill behind the scenes.

Some of the suggested changes to the fault mitigation PR are also mixed in here, since I did it all at once. In addition to the suggested changes, I fixed a bunch of bugs and added a new command to completely wipe the EEPROM. For safety, the wipe EEPROM command needs to be sent twice in a row before it will work, since it should only ever be used during development, or if the EEPROM becomes corrupted.

Finally, I added a flag `STREAM_PHOTO_DATA`, and the corresponding commands to set it. If true, `dataProcessing()` will print the voltage value of each incoming photodiode sample in real time.

### DIFF
--- a/FSW/src/headers/commandHandling.hpp
+++ b/FSW/src/headers/commandHandling.hpp
@@ -51,7 +51,8 @@ namespace commandCode {
         SCRUB_FLASH,                   // start scrubbing the flash memory for errors
 
         // Fault Mitigation
-        CLEAR_ALL_PERSISTENT_DATA,     // clears all persistent data stored in EEPROM
+        WIPE_EEPROM,                   // completely wipes the EEPROM and then resets persistent data
+        RESET_PERSISTENT_DATA,         // resets all persistent data to default values
         ACT_ON_FAULTS_T,               // sets fault action flag so that corrective action is taken
         ACT_ON_FAULTS_F,               // sets fault action flag so that corrective action is not taken
 
--- a/FSW/src/headers/commandHandling.hpp
+++ b/FSW/src/headers/commandHandling.hpp
@@ -39,6 +39,8 @@ namespace commandCode {
         HEATER_OFF,                     // turn heater off, does not override housekeeping heater control
         FORCE_HEATER_ON_T,              // force the heater on, overrrides houskeeping heater control
         FORCE_HEATER_ON_F,              // stop forcing the heater on, giving control back to housekeeping
+        STREAM_TEMPERATURE_T,           // start streaming temperature measurements
+        STREAM_TEMPERATURE_F,           // stop streaming temperature measurements
         CALIBRATE_OPTICS_THERM,         // take new voltage measurement at known temp for optics thermistor
 
         // Command Handling
@@ -57,6 +59,8 @@ namespace commandCode {
         // Fault Mitigation
         WIPE_EEPROM,                   // completely wipes the EEPROM and then resets persistent data
         RESET_PERSISTENT_DATA,         // resets all persistent data to default values
+        SUPPRESS_FAULTS_T,             // stops new faults from being logged
+        SUPPRESS_FAULTS_F,             // allows new faults to be logged
         ACT_ON_FAULTS_T,               // sets fault action flag so that corrective action is taken
         ACT_ON_FAULTS_F,               // sets fault action flag so that corrective action is not taken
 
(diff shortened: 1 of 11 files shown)


FSW/src/headers/commandHandling.hpp
```diff
@@ -22,12 +22,16 @@ namespace commandCode {
...
+        // Mode Change
         ENTER_SAFE_MODE = 1,            // enter safemode state
         ENTER_STANDBY_MODE,             // enter standby state
         ENTER_SUNSET_MODE,              // enter sunset data collection mode
         ENTER_PRE_SUNRISE_MODE,         // enter watch-for-sunset mode
         ENTER_SUNRISE_MODE,             // enter sunrise data collection mode
+
+        // Data Collection
+        STREAM_PHOTO_DATA_T,            // start streaming photodiode data
+        STREAM_PHOTO_DATA_F,            // stop streaming photodiode data
```
jaretanderson, line 34: perfect for testing I love this

FSW/src/util/externalConfig.cpp
```diff
@@ -0,0 +1,31 @@
...
+Event exitMainLoopEvent = Event();
+
+// Data Collection
+volatile bool STREAM_PHOTO_DATA = STREAM_PHOTO_DATA_INIT;
+RecurringEvent dataProcessEvent = RecurringEvent(SAMPLE_PERIOD_MSEC);
+Event saveBufferEvent = Event();
+TimedEvent sunriseTimerEvent = TimedEvent(WINDOW_LENGTH_MSEC);
+TimedEvent sweepTimeoutEvent = TimedEvent(SWEEP_TIMEOUT_MSEC);
+AsyncEvent downlinkEvent = AsyncEvent(MAXFILES);
+AsyncEvent scrubEvent = AsyncEvent(MAXFILES);
```
jaretanderson: I didn't realize you can declare class objects this way. Why this instead of `TimedEvent object(LENGTH)`?

FSW/src/headers/config.hpp
```diff
@@ -6,7 +6,7 @@
  *  sampling rates, etc) that we will want to be able to change
  *  from a single location and include in multiple modules
  * 
- *  declare all objects as static to avoid the error 
+ *  declare all objects as extern to avoid the error 
  *      "multiple definitions of <object>" during compilation
```
jaretanderson: might make sense to explain the relationship between `config.hpp` and `externalConfig.hpp` here? something along the lines of 

`config.hpp` contains all constants that may be useful to change before running the FSW. `externalConfig.hpp` is used to instantiate class objects using `extern` so that they can be modified from any other files in the FSW.

jaretanderson (approved):
It will be really important for us to produce clear documentation on what cases the future devs should put files in `externalConfig.hpp` versus `config.hpp`. Might even suggest changing `externalConfig.hpp` to something like `externalDefinitions.hpp` to make it a little more clear that these things aren't supposed to be messed with in between tests

FSW/src/util/externalConfig.cpp
Vysares: They are equivalent, its just that `TimedEvent object(LENGTH)` is the same syntax as a function declaration, so
`TimedEvent object = TimedEvent(LENGTH)` leaves no ambiguity. 

Not to be confused with `TimedEvent object = new TimedEvent(LENGTH)`, which allocates memory on the heap instead of the stack.

FSW/src/headers/config.hpp
Vysares: changed to:

"declare all non-constant objects as 'extern' and define them in externalDefinitions.cpp to create a single instance shared across all files. Objects not declared as 'extern' will not be linked between files."

I also renamed `externalConfig` to `externalDefinitions` like you suggested, since none of the init values are actually defined there.

JashanChopra (approved):

PR #28, opened 2021-03-24
GUI updates

Vysares:
- Added a button to save output to a file. It will warn you if you try to overwrite a file
- Added a tick box for auto scroll
- Added a button to clear output, which will ask for confirmation
- COM port must now be opened by the user, and commands will not be sent until a connection is established. I had an issue where it tried to connect to my headphone's Bluetooth adapter and scared the hell out of me with super loud static. Now that will only happen if the user opens the wrong port.
- Fixed a few textbox-related bugs
- Changed serial timeout to 50ms
- Set `ACT_ON_NEW_FAULTS` to initialize as false, that way we can still see the faults (command 25 suppresses them completely), but we save the EEPROM life by not writing every fault.

Try it out and see if you can break it; let me know if you can. I want this to be rock solid


FSW/src/headers/config.hpp
```diff
@@ -107,7 +107,7 @@ const bool SUPPRESS_FAULTS_INIT = false;  // whether or not to log new fault occ
 
 // Corrective action
 extern volatile bool ACT_ON_NEW_FAULTS; 
-const bool ACT_ON_NEW_FAULTS_INIT = true; // whether to attempt corrective action when faults are detected
+const bool ACT_ON_NEW_FAULTS_INIT = false; // whether to attempt corrective action when faults are detected
```
jaretanderson, line 112: I like this change for testing, but we would definitely want it to act on faults during something like Thermal right? So we either need to put something in their test procedure to update this to true before the test, or we would want to revert this change before test day. Personally my vote would be put something in the test procedure if you agree

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
+            global portOpen
+            portOpen = True
+            updateLog('Connected to port \"' + portField.get() + '\".')
+            return    
+    updateLog('Port \"' + portField.get() + '\" not found.')
+    
+# save to file #
+def saveFile(): # saves monitor contents to a file
+    fileName = fileNameField.get() + '.txt'
+    filePath = 'saved files\\' + fileNameField.get()
```
jaretanderson: I am not 100% understanding this, does it prompt for the filepath when you press save to file? or just a name?

jaretanderson: Ah I see now that it is in the `saved files` relative path. Maybe update the warning message?

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
+    if (os.path.exists(filePath)): # display warning if file already exists
+        warning = '\"' + fileName + '\" already exists. Do you wish to overwrite it?'
```
jaretanderson: "... already exists in \"saved files\" folder. Do you ...."

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
+        if autoScroll.get():
+            monitor.see(END) 
 
 # scan for serial ports
-def scanPorts():
+def scanPorts(): # finds and lists all available com ports
     availablePorts = list_ports.comports()
     updateLog('==== Found Serial Ports: ====')
     for port in availablePorts:
         updateLog('- ' + port.description)
```
jaretanderson: perhaps print a message saying "no ports" if there are none?
image

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
+            teensy.port = portField.get()
+            teensy.open()
+            global portOpen
+            portOpen = True
+            updateLog('Connected to port \"' + portField.get() + '\".')
+            return    
+    updateLog('Port \"' + portField.get() + '\" not found.')
+    
+# save to file #
+def saveFile(): # saves monitor contents to a file
```
jaretanderson: on my computer it doesn't actually save the files as text files, just as extensionless files. Probably not a huge deal but notably different than what the GUI is telling me is happening

jaretanderson (approved):
if I unplug my teensy from my computer while running the GUI it crashes, which would lead to all of the serial data being lost if the connection was interrupted before the operator saved a file. May be worth implementing some kind of autosave option if that wouldn't be too hard, maybe even just append to one file at set intervals?

This is super cool though and will be really useful for testing, great work

FSW/src/headers/config.hpp
Vysares: I agree, since it can be set via command.

GSW/testUtility/testNS2.pyw
Vysares: Done

GSW/testUtility/testNS2.pyw
Vysares: how do you even have no serial ports?? I'll add that in.

GSW/testUtility/testNS2.pyw
Vysares: same here, I'll look into it

Vysares: works now, it was a dumb mistake

FSW/src/headers/config.hpp
JashanChopra: I vote for a test procedure sending it.

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
 # log frame
 logFrame = LabelFrame(root, text='Log', padx=5, pady=5)
-log = tkScrolledText.ScrolledText(logFrame, state='disabled', width=50, height=20, font=('Consolas',10))
+log = tkScrolledText.ScrolledText(logFrame, state='disabled', width=50, height=20, font=('Consolas',10),  wrap=WORD)
 
 # monitor frame
 monitorFrame = LabelFrame(root, text='NS2 Output', padx=5, pady=5)
 monitor = tkScrolledText.ScrolledText(monitorFrame, state='disabled', width=80, height=30, font=('Consolas',10))
+autoScroll = BooleanVar(value=True)
+check_autoScroll = Checkbutton(monitorFrame, text='Auto Scroll',variable=autoScroll, onvalue=True, offvalue=False)
```
JashanChopra: pedantic missing space before variable

GSW/testUtility/testNS2.pyw
JashanChopra: Lmfao

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
+        if (not overwrite):
+            return
+    # write to file
+    file = open(filePath,'w', newline = '')
+    file.write(monitor.get('1.0','end-1c'))
+    updateLog( 'Output saved to ' + fileName)
+
+# clear output #
+def clearOutput(): # clears the output monitor
+    confirm = tkinter.messagebox.askyesno(title='Warning', message= 'Are you sure you want to clear the output monitor???')
```
JashanChopra: I wonder if a yes/no box would be good on certain commands as well.

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
+fileFrame.grid(row=3, column=0, padx=5, pady=5, sticky=W+E)
+fileNameField.grid(row=0, column=0, sticky=W+E)
+button_saveFile.grid(row=0, column=1, sticky=W+E)
 
 # ==== start application ====
-teensy = serial.Serial(port='COM3', baudrate=9600, timeout=10/1000)
+# configure serial connection
+portOpen = False
+teensy = serial.Serial()
+teensy.baudrate = 19200
```
JashanChopra: What prompted the baud rate change?

GSW/testUtility/testNS2.pyw
```diff
@@ -26,87 +26,148 @@ commandField = Entry(commandFrame, width=30, borderwidth=3, font=('Consolas',10)
...
 commandField.focus() # place cursor in command field
 
-serialThread = threading.Thread(target=readSerial)
-
 while True:
-    readSerial()
-    root.update()   
+    root.update()
+    if portOpen:
+        readSerial()
```
JashanChopra: I would also really like to see an autosave feature. Thermal testing is very stressful, I don't want them to lose data. Can you add an autosave() function that just prints the serial monitor to an autosave file, every ~5 min. You could even just add something simple to the while loop:

now = datetime.now()

if datetime.now() > (now + five_min)
   autosave_file = open(/autosave.txt/, 'w', newline = '')
   autosave_file.write(monitor.get('1.0','end-1c'))
   now = datetime.now()

JashanChopra (approved):

GSW/testUtility/testNS2.pyw
Vysares: I think think the only reason we had it at 9600 was because the Arduino examples had that. Turns out the Teensy can handle a much higher rate. So basically just because we can.

GSW/testUtility/testNS2.pyw
Vysares: Even better, I made it so each line is written to a backup file in real time

GSW/testUtility/testNS2.pyw
Vysares: Yeah it only asks for the name. I added a label to clarify, and updated the warning message

GSW/testUtility/testNS2.pyw
Vysares: I added a flag in the FSW to enable/disable potentially dangerous commands that can cause shutdown or data loss. It will yell at you if you try to execute them without disabling the safeguard.

PR #29, opened 2021-03-25
GUI and command updates

Vysares:
- Fixed an issue where the GUI would sometimes appear blurry.
-  Made the GUI more robust against user error and/or brain failure. It now offers more useful info.
- Added a `Close Serial Port` button, so you don't have to stop the exe to reprogram the Teensy.
- Added a command with code 1 that prints a status report of the payload.
- Opening the serial port now automatically triggers a status report by sending command 1.

Screenshot 2021-03-25 145201


Vysares:
Pyserial throws an exception if the Teensy gets disconnected when the port is open. The software will now catch that exception and close the port instead of just crashing. When the Teensy is reconnected, you can just open the port again without restarting the application.

Screenshot 2021-03-25 152751

FSW/src/headers/commandHandling.hpp
```diff
@@ -18,13 +18,17 @@ bool checkIfCommandAllowed(int command);
...
 
 namespace commandCode {
     // command codes are wrapped in a namespace so they are not global
     enum Code { // all possible commands
         // commands are 1 indexed so that the default initializer can be used for the command queue
+        // Ping
+        INFO = 1,                       // Triggers a readback of system data
+
         // Mode Change
-        ENTER_SAFE_MODE = 1,            // enter safemode state
```
JashanChopra, line 27: Make sure to update the docs if all these command numbers shifted. We ran into that today. Putting in the prevention from sending dangerous commands saved us, so good change there.

GSW/TestUtility/Source/testNS2.py
```diff
@@ -56,27 +62,37 @@ def updateLog(text): # writes text to the log
...
     if command:
-        teensy.write(bytes(command, 'utf-8'))
-        updateLog('Command Sent: ' + command)
+        sendCommand(command)
         commandField.delete(0, END)
 
 # read serial #
 def readSerial(): # reads incoming communication from NS2 via the open serial port
-    data = str(teensy.readline())[2:-1]
+    try:
```
JashanChopra, line 89: try except statements make me so happy

GSW/TestUtility/Source/testNS2.py
```diff
@@ -98,21 +114,31 @@ def scanPorts(): # finds and lists all available com ports
         updateLogPlain('- ' + port.description)
         foundPorts = True
     if not foundPorts:
-        updateLog('No ports detected.')
+        updateLogPlain('No ports found. Is the Teensy connected?')
                 
 # open serial port #
 def openSerialPort(event=None): # opens a new serial connection if port exists
+    if not re.match(r"COM[0-9]+", portField.get()):
```
JashanChopra, line 121: Was this in response to a specific error, or just checking the regex here for the hell of it?

JashanChopra (approved):

FSW/src/headers/commandHandling.hpp
Vysares: Will do. I've been holding off since I've just been adding so many commands, but now that we are starting to test with the hardware I'll make sure to keep it up to date

GSW/TestUtility/Source/testNS2.py
Vysares: The `openSerialPort` command just looks for any port with a description *containing* the value in the entry field, that means that if you enter something like 'C' in the field, it will try to open every available port.

Vysares: *would try to open every available port. The regex will prevent that now.

FSW/src/headers/commandHandling.hpp
```diff
@@ -71,6 +78,7 @@ namespace commandCode {
         EXIT_MAIN_LOOP,                // exits the main loop
 
         // End of list
+        SELF_DESTRUCT,
```
jaretanderson, line 81: can't believe I forgot to add this functionality earlier, this will be really nice to have

jaretanderson (approved):
Looks great, tons of good fixes here

PR #35, opened 2021-03-30
Fault rework

Vysares:
I went in and did a fairly in depth rework of the fault handling logic. I think it makes a lot more sense now.

- The `SUPPRESS_FAULTS` flag and its corresponding command pair no longer prevents fault logging. Instead it just prevents messages from being printed each time a fault is logged.
- `ACT_ON_FAULTS` will now only act on faults that are logged _after_ it is enabled. This prevents multiple conflicting faults all calling for corrective action (like too hot and too cold at the same time). 
- Faults now call for action every time they are logged as long as `ACT_ON_FAULTS` is true. But action will only be taken if necessary to avoid repeating actions that do literally nothing (like setting a flag to false that is already false).
- Corrective actions no longer use the command system to avoid conflicting with the command queuing functionality.
- Added a new global flag `SAVE_FAULTS_TO_EEPROM` that does exactly what it says. This function was previously coupled to `ACT_ON_FAULTS` for no good reason. We may want to act on faults for testing without stressing the EEPROM needlessly. We only get 1080 bytes!
- Each time persistent data is written to EEPROM its address is shifted so we cycle through each block of memory sequentially. This will theoretically extend the memory life by ~12 times. A new function `seekEEPROM()` finds the address of the most recent write so data can be read from the EEPROM on startup.
- Added more info to the `INFO` command.

Finally, I've been running the same instance of the GUI exe all day to test over a dozen FSW builds, all without without a single crash or hangup. That's a good sign.


FSW/src/util/commandHandling.cpp
```diff
@@ -430,13 +440,28 @@ void executeCommand(int command) {
...
+            break;
+        case STANDBY_MODE: Serial.println("Standby");
+            break;
+        case SUNSET_MODE: Serial.println("Sunset");
+            break;
+        case PRE_SUNRISE_MODE: Serial.println("Watching for Sunrise");
+            break;
+        case SUNRISE_MODE: Serial.println("Sunrise");
+            break;
+        default: Serial.println("Mode Not Recognized!");
```
JashanChopra, line 453: If I'm the test operator and I see this, I'd be pretty scared. Maybe switch to safe mode and also print that?

FSW/src/util/faultManager.cpp
```diff
@@ -100,51 +102,63 @@ void feedWD() {
...
+
+            case faultCode::UNEXPECTED_RESTART:
+                scienceMode.setMode(SAFE_MODE);
+                Serial.println("Corrective Action Taken - Entering Safe Mode.");
+                Serial.println("(Unexpected restart!)");
+                break;
+
+            // temp too hot
+            case faultCode::ANALOG_TOO_HOT:
+            case faultCode::DIGITAL_TOO_HOT:
```
JashanChopra, line 134: I think making sure the heater is off is a valid corrective action here.

FSW/src/util/faultManager.cpp
```diff
@@ -100,51 +102,63 @@ void feedWD() {
...
+            case faultCode::OPTICS_TOO_HOT:
+
+            // temp too cold
+            case faultCode::ANALOG_TOO_COLD:
+            case faultCode::DIGITAL_TOO_COLD:
+            case faultCode::OPTICS_TOO_COLD:
+                if (HEATER_OVERRIDE) {
+                    HEATER_OVERRIDE = false;
+                    Serial.println("Corrective Action Taken - Automatic heater control re-enabled.");
+                    Serial.println("(One or more temperatures out of acceptable range!)");
```
JashanChopra: One or more temperatures **below** acceptable range

JashanChopra (approved):

FSW/src/headers/faultManager.hpp
```diff
@@ -37,10 +34,10 @@ namespace faultCode {
 
 struct FaultReport {
     uint8_t occurrences = 0;
-    uint8_t pendingAction = 0; 
     uint16_t startNum = 1;
     uint32_t timestamp = 0;
-    static const int MEMSIZE = 8;
+    bool pendingAction = 0; // not written to EEPROM
```
jaretanderson: maybe stick with the `false` convention instead of `0` if this is a bool?

FSW/src/headers/faultManager.hpp
```diff
@@ -24,9 +24,6 @@ namespace faultCode {
         DIGITAL_TOO_COLD,
         OPTICS_TOO_HOT,
         OPTICS_TOO_COLD,
-
-        // Power
-        DREG_OUT_OF_RANGE,      // Digital regulator voltage is out of normal range
```
jaretanderson, line 29: testing on my local machine revealed that this variable is still used in `housekeeping.cpp` so you will have to take it out of there for the code to compile

jaretanderson (changes requested):
the code doesn't currently compile on my computer so I would like to get it to that state at least before merging

jaretanderson (approved):

FSW/src/headers/faultManager.hpp
Vysares: good catch

FSW/src/util/commandHandling.cpp
```diff
@@ -255,36 +255,21 @@ void executeCommand(int command) {
...
             break;
 
-        case commandCode::STREAM_PHOTO_SPI_T:
-            STREAM_PHOTO_SPI = true;
-            Serial.println("Command Executed - Transmitting photodiode voltages read by SPI in real time.");
-            Serial.println("PHOTO_SPI | time (ms) | photodiode voltage (V)");
+        case commandCode::STREAM_PHOTO_T:
+            STREAM_PHOTO = true;
+            Serial.println("Command Executed - Transmitting photodiode voltages read in real time.");
+            Serial.println("PHOTO_DIR | time (ms) | SPI voltage (V) | Pin voltage (V)");
```
jaretanderson: maybe just have the prefix be `PHOTO` since we are streaming both?

FSW/src/util/commandHandling.cpp
```diff
@@ -255,36 +255,21 @@ void executeCommand(int command) {
...
-            Serial.println("Command Executed - Printing next photdiode sample from SPI.");
-            break;
-
-        case commandCode::PRINT_PHOTO_DIR_SINGLE:
-            printEventDir.invoke();
-            Serial.println("Command Executed - Printing next photdiode sample from Teensy ADC.");
+        case commandCode::PRINT_PHOTO_SINGLE:
+            printPhotoEvent.invoke();
+            Serial.println("Command Executed - Printing next photdiode sample.");
+            Serial.println("PHOTO_DIR | time (ms) | SPI voltage (V) | Pin voltage (V)");
```
jaretanderson: same `PHOTO` prefix comment

FSW/src/util/dataCollection.cpp
```diff
@@ -75,21 +75,16 @@ uint16_t dataProcessing() {
      */
 
     // print the voltage value (for testing)
-    if (printEventSPI.checkInvoked() || STREAM_PHOTO_SPI) {
+    if (printPhotoEvent.checkInvoked() || STREAM_PHOTO) {
         float spiVoltage = photodiode16 * ADC_VOLTAGE_RES;
+        float directVoltage = analogRead(PIN_PHOTO) * TEENSY_VOLTAGE_RES;
         Serial.print("PHOTO_SPI, ");
```
jaretanderson: same `PHOTO` prefix comment

jaretanderson (approved):

FSW/src/util/faultManager.cpp
Vysares: In the unlikely event that one temperature is too hot and another is too cold the heater would continuously toggle. Passing control over to housekeeping if _any_ temperature is too hot or too cold will automatically prioritize the optics temp and set the heater accordingly. There is a margin between the heater set temps and the temperatures required to log these faults, so automatic control accomplishes the same thing as setting the heater off directly. I think that scenario requires some more thought though

FSW/src/util/commandHandling.cpp
Vysares: oh whoops

FSW/src/util/commandHandling.cpp
Vysares: `setMode()` will already default to safe mode if the mode isn't recognized, so I just put that in for the sake of having a default case really

